PR #2, opened 2019-04-26
feat: add do-dont component

alisonjoseph:
- Add DoDontExample
- Add ImageComponent
- Update ibm-- to bx--
- Update AnchorLinks
- **REMOVE OFFSET CLASSES** 🎉


gatsby-theme-carbon/src/components/DoDontExample/do-dont-example.scss
```diff
@@ -0,0 +1,156 @@
...
+.#{$prefix}--example__icon--incorrect path[dataIconPath="inner-path"] {
+  fill: $carbon--white-0;
+  opacity: 1;
+}
+
+.#{$prefix}--example__caption {
+  padding: $spacing-03 0 0 0;
+}
+
+p.#{$prefix}--example__title {
```
vpicone: do we need to specify the element for these guys? It think the class names are solid on their own.

gatsby-theme-carbon/src/components/AnchorLinks/anchor-links.scss
```diff
@@ -0,0 +1,67 @@
...
+  width: auto;
+  margin-bottom: 0;
+  margin-top: $spacing-05;
+
+  @include carbon--breakpoint("md") {
+    margin-bottom: $spacing-07;
+  }
+}
+
+.#{$prefix}--anchor-links .markdown-ul li {
```
vpicone: This is a good example of where sass nesting can fail us. If the user shadows the Ul component, the AnchorLink component will break since the `markdown-ul` won't exist.

We should use a new className for the li over nesting as we can't depend on the classNames of any other component existing outside of the current component.

vpicone:
Disabled the ImageComponent nav link, I think it needs a lot of work. 

Zoom, overlay, and captions are broken. I think we need to get rid of this sort of syntax in favor of what I did with the `AnchorLink` component.

It's confusing to look at, requires some really hacky React and prone to breaking with whitespace issues. 

```
<DoDontExample label="label text">

![Alt text](images/Light_theme_02.png)

</DoDontExample>
```

PR #26, opened 2019-05-02
feat: caption

### LINKED ISSUE #17: Caption component & image adjustments
Captions on the Carbon site are built into the image component. With the removal of the need for the offset class, we don't need the image component as a default. We can just place inside our `<Col>` component to have it display at the correct width. 

However, this means we need a separate Caption component

`<Caption>This is the caption</Caption>`
or
`<Caption className="my-custom-class">This is the caption</Caption>`

html output should be
```
<div class="bx--col-lg-4 bx--col-md-4">
    <p class="bx--caption my-custom-class">Glyph usage in Buttons</p>
</div>
```
sass
```
padding-top: $spacing-03;
margin-bottom: $spacing-06;
margin-top: -$spacing-06; // to account for default spacing below images
```

- Update default image styles to have margin-bottom: `$spacing-06`

alisonjoseph:
closes #17

### DIFF
--- a/packages/example/src/data/nav-items.yaml
+++ b/packages/example/src/data/nav-items.yaml
@@ -4,6 +4,8 @@
       path: /component-examples/demo
     - title: DoDontExample
       path: /component-examples/DoDontExample
+    - title: Caption
+      path: /component-examples/Caption
     # - title: ImageComponent
     # path: /component-examples/ImageComponent
 - title: Contributing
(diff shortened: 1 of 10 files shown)


vpicone:
What if instead of depending on the row/column relationship, we have a fullWidth prop. So the default class has width 100 with a separate class that apply a the responsive styles. Unless they send the fullWidth prop, we add on the responsive class as well. 

Then we can reuse that responsive class in other components as well.

alisonjoseph:
I'm worried that is overcomplicating what should be a super simple component. Since the goal is to make this easy for non-devs to use these components in markdown.

vpicone:
I think just the one Boolean prop would be worth the flexibility. If someone wanted to use a caption in their own component or container they would be out of luck.

Issue #21, opened 2019-04-30
Card components (the component formerly known as ClickableTile)


alisonjoseph:
Create a reusable tile component. Determine which options are needed, look at IDL, ibm.com/design, Carbon, IDR, AI sites to make sure it covers all of them.

alisonjoseph:
![Screen Shot 2019-04-30 at 3 29 12 PM](https://user-images.githubusercontent.com/2753488/56991368-1b8f0880-6b5d-11e9-8cf0-c361466353fb.png)
![Screen Shot 2019-04-30 at 3 31 29 PM](https://user-images.githubusercontent.com/2753488/56991369-1b8f0880-6b5d-11e9-975b-adac24c51499.png)
![Screen Shot 2019-04-30 at 3 31 24 PM](https://user-images.githubusercontent.com/2753488/56991370-1c279f00-6b5d-11e9-88f1-1f60ebe754b9.png)
![Screen Shot 2019-04-30 at 3 31 17 PM](https://user-images.githubusercontent.com/2753488/56991371-1c279f00-6b5d-11e9-8ced-375a043b13ec.png)
![Screen Shot 2019-04-30 at 3 30 27 PM](https://user-images.githubusercontent.com/2753488/56991374-1c279f00-6b5d-11e9-9fd1-465a89cc0943.png)
![Screen Shot 2019-04-30 at 3 30 00 PM](https://user-images.githubusercontent.com/2753488/56991375-1c279f00-6b5d-11e9-830a-87cfb1feb5c7.png)
![Screen Shot 2019-04-30 at 3 29 50 PM](https://user-images.githubusercontent.com/2753488/56991376-1cc03580-6b5d-11e9-8cfd-e95ddde47309.png)
![Screen Shot 2019-04-30 at 3 29 34 PM](https://user-images.githubusercontent.com/2753488/56991377-1cc03580-6b5d-11e9-9803-b517f842d5d7.png)
![Screen Shot 2019-04-30 at 3 29 30 PM](https://user-images.githubusercontent.com/2753488/56991379-1cc03580-6b5d-11e9-86fb-43321fac249b.png)
![Screen Shot 2019-04-30 at 3 29 21 PM](https://user-images.githubusercontent.com/2753488/56991380-1cc03580-6b5d-11e9-9b67-932ae22b5365.png)

sstrubberg:
[...]

PR #27, opened 2019-05-02
feat: add ResourceCard component

alisonjoseph:
`ResourceCard` is the simple small clickable button card, the other variations will be separate components. 

image

variations will be
dark
disabled
aspectRatio

todo

- [x] make it smart enough to differentiate between link and href
- [x] create a wrapper component/determine a way to do the border/outline between tiles in groups 😅 
- [x] add additional aspect ratios

ref #21


alisonjoseph:
@vpicone would love your thoughts on the `<Row className="resource-tile-group">` group class name I'm having to add to get the correct styles for groups of tiles. Since this is super specific to this layout. Was thinking we provide this, and in the docs show as an example of how they can add their own class and styles if they need a different layout? 🤔 Unfortunately, the design isn't just 1px between tiles, its actually a darker gray than the background so there isn't an easy way to know whether to set the top or left borders.

PR #39, opened 2019-05-09
feat: add anchor tags and smooth scroll

vpicone:
This PR adds an `AutolinkHeader` component is used in h1, h2 and h3 MDX components to add auto generated anchor tags and heading ids.

This approach means we have total control over the links and the tag element when compared to `gatsby-remark-autolink-headers`.


alisonjoseph:
Looking good! I think we might also need this for h4s (we have it linking h4s on Carbon) but we should double check w/ @jeanservaas on that

packages/gatsby-theme-carbon/src/components/markdown/AutolinkHeader.js
```diff
@@ -0,0 +1,59 @@
+import React from 'react';
+import GHSlugger from 'github-slugger';
+import Link from '@carbon/icons-react/lib/link/20';
+import { css } from '@emotion/core';
+import styled from '@emotion/styled';
+
+const slugger = new GHSlugger();
+
+const Anchor = styled.a`
```
alisonjoseph: Can we put these styles in a sass file 😇

vpicone: className toggling is for squares!

vpicone: ( I'll check it out :) )

vpicone:
I switched the AutolinkHeader to use what I think would be our ideal sass setup [for components](https://github.com/carbon-design-system/gatsby-theme-carbon/tree/675dadd69d019801947b3289c7dd997c21b6085e/packages/gatsby-theme-carbon/src/components/AutolinkHeader
).

Using the module approach ensures we won’t have any name collisions between the devs and our styles which would eliminate the need for prefixing in both css and js files. It also prevents the possibility of smelly global patterns and  inter-component collisions within our own app (eliminates the need for contrived BEM-ish class names)

packages/gatsby-theme-carbon/src/components/AutolinkHeader/AutolinkHeader.js
```diff
@@ -0,0 +1,36 @@
+import React from 'react';
+import GHSlugger from 'github-slugger';
+import Link from '@carbon/icons-react/lib/link/20';
+import cx from 'classnames';
+
+import { header, anchor } from './AutolinkHeader.module.scss';
```
vpicone: You can destructure the class names from the sass module.

`hypenated-class` would become `hyphenatedClass`

you can also just do `import styles from 'x.module.scss'` and then use styles.anchor etc.

packages/gatsby-theme-carbon/src/components/AutolinkHeader/AutolinkHeader.module.scss
```diff
@@ -0,0 +1,25 @@
+@import '~carbon-components/scss/globals/scss/vars';
```
vpicone: Only importing the necessary sass. It's important this sass is variables only so it doesn’t generate css on every module.

packages/gatsby-theme-carbon/src/components/markdown/AutolinkHeader.js
vpicone: > Looking good! I think we might also need this for h4s (we have it linking h4s on Carbon) but we should double check w/ @jeanservaas on that

She said h1-h3 is actually preferred.

PR #62, opened 2019-05-16
feat: add sticky tab feature

vpicone:
TODO:

- [X] Implement sticky behavior on pages with tabs
- [X] Global header hide/reveal on scroll for desktop

Changes:
- Refactor PageHeader and PageTabs to use scss modules
- Adjust z-index for overlay, nav, and site-switcher to match spec


vpicone:
@alisonjoseph @jeanservaas I wanted to test the darker type color for inactive tabs like you see on the IDL. Wondering what you think about it vs. the carbon website version where they're all white all the time.

screen

jeanservaas:
LGTM but are we still doing the header hide on scroll in desktop?

jeanservaas:
Like darker type color for inactive btw

vpicone:
> LGTM but are we still doing the header hide on scroll in desktop?

Yah, it’s in my todo list at the top

vpicone:
> the alignment of the type within the tabs looks a little off maybe? compared with IDL site:
> Screen Shot 2019-05-16 at 1 36 07 PM
> 
> other than that, looks good! this current state of always keeping the global header visible is also a good option.

Unfortunately the IDL doesn't even align with itself sometimes 🤣
Screen Shot 2019-05-16 at 2 54 03 PM

I updated them across the board to account for the border-top though thanks!

PR #79, opened 2019-05-20
Consider opening ResourceLinks in new tab

josefaidt:
Consider opening outbound links in a new tab.

## Changes

- add `target="_blank"` to open in a new tab
- add `rel="noopener noreferrer"` for security and performance


josefaidt:
Added an additional check to only add extra props if link href includes `http`

packages/gatsby-theme-carbon/src/components/LeftNav/ResourceLinks.js
```diff
@@ -7,18 +7,23 @@ const LeftNavResourceLinks = ({ links }) =>
...
-        >
-          {link.title}
-        </SideNavLink>
-      ))}
+      {links.map((link, i) => {
+        const linkProps = !link.href.includes('http')
+          ? { element: Link }
+          : { element: 'a', target: '_blank', rel: 'noopener noreferral' }
+        return (
+          <StyledLink
```
alisonjoseph: Did you mean to change this to `<StyledLink>` ? getting a build error
```
15 |           <StyledLink
     |           ^
  16 |             key={i}
  17 |             style={{ marginTop: i === 0 ? '1rem' : 0 }}
  18 |             icon={<LaunchIcon />}


  WebpackError: ReferenceError: StyledLink is not defined
  
  - ResourceLinks.js:15 
```

josefaidt: oof, I was wrapping it to change some styling. Changed it back

josefaidt:
That is a good point. I could just revert it and spread props into the link component

vpicone:
I've always been conflicted about this, If I wanted to open it in a new tab I'd middle/cmd+click. I generally don't mess with the default anchor link behavior on my own projecrts but could go either way here.

vpicone:
Also, thanks so much for chipping in @josefaidt! Excited to have you playing with the project.

josefaidt:
I think I may have misunderstood the "Resource Links" purpose initially, which is what prompted this PR. I have no ties to it, but I'm re-rolling the component myself to get the functionality. Would it be more beneficial to add a prop to the link object that enables the ability of opening to a new tab? 
Something like:

```js
const links = [
  {
    title: 'Some Title'
    href: 'https://google.com'
    outbound: true,
  },
]
```

We could also use that flag to set the appropriate icon. Currently I'm wrapping the `SideNavLink` component to render the icon (this is due to Carbon scss targeting only the last two items in the SideNav), using:

```js
const StyledLink = styled(SideNavLink)`
  flex-direction: row-reverse;
  justify-content: space-between;

  & > div.bx--side-nav__icon {
    display: block !important;
    flex: 0 !important;
  }
`
```

vpicone:
What if we instead of taking an opinionated stance on opening new tabs we let the developer supply additional link props when they define their links array. 

Then we spread those values into the links. Would allow for things like `aria-label` as well.

So shadowing would look like:

`example/src/gatsby-theme-carbon/components/LeftNav/ResourceLinks.js`

```jsx
import React from 'react';
import ResourceLinks from 'gatsby-theme-carbon/src/components/LeftNav/ResourceLinks';

const links = [
  { title: 'Github', href: 'https://github.com', ['aria-label']: 'link to github' },
  { title: 'Storybook', href: 'https://react.carbondesignsystem.com' },
];

const CustomResources = () => <ResourceLinks links={links} />;

export default CustomResources;
```

josefaidt:
Something like this?

```js
const links = [
  {
    title: 'Slack',
    href: 'https://ibm-gbs-lk.slack.com',
    target: '_blank',
    rel: 'noopener noreferral',
  },
  {
    title: 'Github',
    href: 'https://github.ibm.com/',
    target: '_blank',
    rel: 'noopener noreferral',
  },
]

const CustomResources = () => (
  <>
    <hr className="bx--side-nav__divider" />
    {links.map((link, i) => (
      <SideNavLink
        key={i}
        style={{ marginTop: i === 0 ? '1rem' : 0 }}
        icon={<LaunchIcon />}
        href={link.href}
        className="bx--side-nav--website-link"
        element={!link.href.includes('http') ? Link : 'a'}
        {...link}
      >
        {link.title}
      </SideNavLink>
    ))}
  </>
)
```

This actually adds the `title` attribute as well, which may be an unwanted side effect?

image

vpicone:
@josefaidt HA looks like an unfortunate consequence of a poor naming choice. We could change the `title` property to `text` or add a title="" after the spread.

I'd also be okay with a `shouldOpenNewTabs` prop to resource links that enables the new tab behavior for all outgoing links. Probably better developer experience than specifying per link.

josefaidt:
I reverted a few of the changes and added `shouldOpenNewTab`, not sure if this is the implementation you were thinking of. 

```js
const LeftNavResourceLinks = ({ links }) =>
  links ? (
    <>
      <hr className="bx--side-nav__divider" />
      {links.map((link, i) => {
        const additionalProps = !link.shouldOpenNewTab ? '' : ({
          element: 'a',
          target: '_blank',
          rel: 'noopener noreferral'
        })
        return (
          <SideNavLink
            key={i}
            style={{ marginTop: i === 0 ? '1rem' : 0 }}
            icon={<LaunchIcon />}
            className="bx--side-nav--website-link"
            element={!link.href.includes('http') ? Link : 'a'}
            {...additionalProps}
          >
            {link.text}
          </SideNavLink>
      )})}
    </>
  ) : null;
```

Were you thinking add `shouldOpenNewTabs` in the ResourceLink component? To enable *all* links in the array to open in a new tab?

josefaidt:
This prompts another decent question I have, do yall want to display the launch icon if the link is set to open in a new tab? I had to do some additional wrapping and styling to mitigate the UI Shell styling, which targets only the last two links on the navbar

```js
const StyledLink = styled(SideNavLink)`
  flex-direction: row-reverse;
  justify-content: space-between;

  & > div.bx--side-nav__icon {
    display: block !important;
    flex: 0 !important;
  }
`
```

vpicone:
@josefaidt wrestling with inflexible UI shell styling has been the bane of this project. It's already been hacked to be way different than the actual component behavior. I'd like to avoid over-riding further and consider re-writing the component with the sass+css module aproach we're taking elsewhere.

vpicone:
>Were you thinking add shouldOpenNewTabs in the ResourceLink component? To enable all links in the array to open in a new tab?

@josefaidt yeah sorry, the per-link implementation looks really gnarley

vpicone:
I think the primary issue here is an inflexible API coming from the links prop. It should just be children so the developer can pass in whatever they want. Presumably sidenav links with or without the props they prefer.

vpicone:
@josefaidt I feel like I took this PR off course haha. If you want to make external links open in a new tab I'm cool with that. We can investigate making the API more flexible if the LeftNav get's re-written.

josefaidt:
Hi Vince! Apologies for taking a good minute to respond, I've gotten caught up in some new training. Revisiting this I've added a check to declare that all links provided beginning with `http` will open in a new tab. This **assumes** such links are outbound, and currently will not always have an "external" identifier due to the nature of the current UI shell.

vpicone:
@josefaidt no worries! dumb question: have you tried just supplying those props no matter what? I know the Link element from gatsby can function as a regular link as well, just wondering if the check is necessary or we can just always throw them on.

josefaidt:
I haven't tried that, and the reason being that the internal links should open in the same tab imo. My thought process for the link usability:

|          | Element | Open in new tab | include LaunchIcon |
| -------- | ------- | --------------- | ------------------ |
| Internal | `Link`  | false           | false              |
| External | `a`     | true            | true               |

I believe maybe we don't *need* the new tab, but maybe if we could define whether or not to display the launch icon to ultimately differentiate internal links from external. This may go back on the design of the UI shell.

vpicone:
Nice, I think we're going to be re-writing a the left nav since it already has some hacks around shell. You should fly out to Austin so we can work on it!

josefaidt:
Ha I'd love to come out and help, don't tempt me! I could add some checks to add the icon, we'd just have to use that wrapped component to override the defaults of the UI Shell side navigation styles. Let me see what I can come up with

josefaidt:
Ok, I don't think you're gonna like this as its a bit brash, but I got the logic to use different, wrapped links depending on the user-provided link URL. This implementation bases the launchIcon render solely off of whether the link provided is prefixed with `http`, drawing an assumption that it will be an outbound link. (it can be cleaned up but it works)

```js
const StyledOutboundLink = styled(SideNavLink)`
  flex-direction: row-reverse;
  justify-content: space-between;

  & > div.bx--side-nav__icon {
    display: block !important;
    flex: 0 !important;
  }
`
const StyledInternalLink = styled(SideNavLink)`
  flex-direction: row !important;
  & > div.bx--side-nav__icon {
    display: none !important;
  }
`

const LeftNavResourceLinks = ({ links }) =>
  links ? (
    <>
      <hr className="bx--side-nav__divider" />
      {links.map((link, i) => {
        const NavElement = !link.href.includes('http') ? StyledInternalLink : StyledOutboundLink
        const Link = !link.href.includes('http') ? ({
          element: Link,
          href: link.href
        }) : ({
          element: 'a',
          target: '_blank',
          rel: 'noopener noreferral',
          href: link.href
        })
        return (
          <NavElement
            key={i}
            style={{ marginTop: i === 0 ? '1rem' : 0 }}
            icon={<LaunchIcon />}
            className="bx--side-nav--website-link"
            {...Link}
          >
            {link.title}
          </NavElement>
        )
      })}
    </>
  ) : null;
```

Modifying the supplied links in the example app, 

```js
const links = [
  { title: 'Github', href: 'https://github.com' },
  { title: 'Resources', href: '/resources' },
  { title: 'Resources', href: '/resources' },
  { title: 'Resources', href: '/resources' },
  { title: 'Resources', href: '/resources' },
  { title: 'Storybook', href: 'https://react.carbondesignsystem.com' },
  { title: 'Resources', href: '/resources' },
  { title: 'Resources', href: '/resources' },
];
```

This will be rendered as,

image

vpicone:
Closing this in favor of https://github.com/carbon-design-system/gatsby-theme-carbon/pull/196

Thanks for helping me think through this!

PR #80, opened 2019-05-20
[WIP] feat: update code snippet to render live editor

jendowns:
## Proposal 

MDX comes with the ability to add a live code editor: https://mdxjs.com/guides/live-code/#live-code

This is something the Security team is super interested in, but I believe that due to the structure of this project, the live editor option needs to be introduced in *this* library so that we can use it.

I propose adding a variation to the `Code` component. By default, code in this theme will display with Carbon's `CodeSnippet` component. If the user adds `.mdx live=true` to the triple tick marks for a code block, then that renders a live code editor.

I've included an example of both Carbon's `CodeSnippet` and the new editor on the demo page.

### 1. Theming for the live editor?
Right now I'm just importing a default theme like so:
```
import theme from 'prism-react-renderer/themes/dracula';
```

Can we make a Carbon theme object that would work instead? I can take a stab at this, if you like?

### 2. Scope needs to be handled more elegantly
The `LivePreview` component only imports `React` by default... so in order to preview a component with `LivePreview`, you need to pass that component in the `scope` prop like this:

```
import { Button } from 'carbon-components-react';

const scope = { Button };

<LiveProvider scope={scope} ... />
```

But there's probably something more elegant that can be done, especially from the context of a consumer who would need to be able to control what the scope should be.

### 3. How to style errors?

If you look at the demo page and try to use some invalid code (or anything out of the `Button` component scope) in the editor, you'll see an unstyled error message appear below the editor. How should that look?


alisonjoseph:
This is awesome, I can definitely see where this could come in handy, especially for component examples! 

I think we definitely need some input from design @jeanservaas before anything like this gets merged into the core theme though.

vpicone:
I think the code snippet implementation is due for some increased flexibility and functionality. This is an awesome start, thanks so much for putting it together! Some notes,

1. Converting our current prism stylesheet to JSON so it can be shared between both live and regular versions of the code would be the way to go. There's surprisingly few rules. https://github.com/carbon-design-system/carbon-website/blob/master/src/styles/prism.css

2. We'll have to think more about this, since it could change between sites. It might be able to work as a [theme option](https://www.gatsbyjs.org/docs/themes/api-reference/#configuration) but not positive.

3. Definitely agree we'll need a design touch on the implementation. Unfortunately just eating the error isn't really an option from a ux perspective I think. Also, If you wanted to get a head start on styling the preview area I think using the code example components on the carbon website would be a great start: https://www.carbondesignsystem.com/components/button/code

@jeanservaas mentioned that forgoing the `CodeSnippet` component could be an option. We discussed the fact that being able to select all the text contained within a snippet took away some of the need for the copy button and also that _good_ code snippets likely shouldn't require a show-more functionality and instead might be better broken up. 

With that in mind, we probably want to more closely stick to the example on the MDX website for live highlighting, particularly as it pertains to non-live code snippets.

jendowns:
Thanks @vpicone I can get started on those points! 👍 

One question about your comment:
> Unfortunately just eating the error isn't really an option from a ux perspective I think.

Are you referring to the `<LiveError />`? If you are implying it should be removed... IMO there's a huge UX value in keeping that and styling it nicely, because it helps people using the live editor know if they have a syntax error that is preventing the preview from rendering correctly.

vpicone:
@jendowns I was actually suggesting the opposite! Until we get that styled, we can't ship this as default, since leaving it out all together is a non-starter.

vpicone:
@jendowns It's really starting to come together great stuff!

vpicone:
Closing in favor of #242 for code component overhaul and #250 to explore the live editing aspect

PR #94, opened 2019-05-22
feat: disable global search

### LINKED ISSUE #93: Global search should be disabled by default
The global search is really only helpful for documentation sites, it should be disabled by default but accessible as an option for sites wanting to enable search functionality.

vpicone:
closes #93 

This disables the GlobalSearch component by default. If you'd like to implement search functionality, you'll need to follow these two steps:

1. set the isSearchEnabled theme option to true
```js
  __experimentalThemes: [
    {
      options: {
        isSearchEnabled: true
      },
    },
  ],
```

2. Shadow the example search implementation at `/src/util/hooks/useSearch.js`

```jsx
import { useEffect } from 'react';
const useAlgoliaSearch = () => {
  // ...
};

export default useAlgoliaSearch;
```

The example `useSearch` hook demonstrates implementing search with [Algolia](https://www.algolia.com/). Algolia is free for open source libraries. You can shadow this hook and replace it with your Algolia credentials or a library of your choosing.


vpicone:
Yeah this would go in the configuration/shadowing section of the docs.

Issue #99, opened 2019-05-23
global: body copy width should be 8 columns wide minus 32px


mjabbink:
@vpicone @alisonjoseph Double check with Jeannie but I think she created this ish in Carbon site and has not yet moved here.

alisonjoseph:
She created the issue both places https://github.com/carbon-design-system/gatsby-theme-carbon/issues/55

mjabbink:
@alisonjoseph @vpicone Did this get done? It’s not reflected in the Gatsby them site yet.

alisonjoseph:
I think it's in an open PR that hasn't gotten merged, I can push just this update in though now.

mjabbink:
:+1:

mjabbink:
@alisonjoseph @vpicone The body copy expands the full 8 columns. It should be minus 32px so it does not go all the way to the far edge of the 8th column.

mjabbink:
What’s the status on this ish?

mjabbink:
<img width="860" alt="Screen Shot 2019-06-04 at 8 33 36 AM" src="https://user-images.githubusercontent.com/32881239/58879456-b3739b00-86a3-11e9-9d6e-b8d86dc6786d.png">

alisonjoseph:
Where is the above screenshot from? I checked and this seems ok in the actual theme. 
![Screen Shot 2019-06-10 at 12 50 08 PM](https://user-images.githubusercontent.com/2753488/59215225-546bc580-8b7e-11e9-9329-72fab0f10c29.png)

mjabbink:
The above one is from brand Center 
https://www.ibm.com/brand/expression

alisonjoseph:
got it, looks like its too much padding now. will look into it
![Screen Shot 2019-06-10 at 1 15 55 PM](https://user-images.githubusercontent.com/2753488/59216749-ecb77980-8b81-11e9-9c33-d3d057ecaefb.png)

mjabbink:
Does look like too much now.

👍

alisonjoseph:
got a fix in this PR https://github.com/carbon-design-system/gatsby-theme-carbon/pull/156
![Screen Shot 2019-06-10 at 1 28 12 PM](https://user-images.githubusercontent.com/2753488/59217484-a6fbb080-8b83-11e9-8ae9-87d351710ccc.png)

PR #123, opened 2019-05-29
feat: add all components globally

alisonjoseph:


packages/gatsby-theme-carbon/src/components/MDXProvider.js
```diff
@@ -3,6 +3,22 @@ import { MDXProvider as Provider } from '@mdx-js/react';
...
+  Row,
+  Column,
+  Grid,
+  Caption,
+  ResourceCard,
+  ArticleCard,
+  Aside,
+  FeatureCard,
+  ImageCard,
+} from 'gatsby-theme-carbon';
```
vpicone: I think this might be the issue. You're attempting to import from `gatsby-theme-carbon` from within `gatasby-theme-carbon`. Maybe try just using relative imports.

alisonjoseph: 🤦‍♂ thanks. this is what happens when I'm switching between three different projects in one day. lol

alisonjoseph:
don't merge until #124 is merged

vpicone:
@alisonjoseph should we change `Tabs.mdx` to use the global instead of importing? Either way, looks good to me.

alisonjoseph:
@vpicone just updated the mdx page 👍

alisonjoseph:
also added the fix for https://github.com/carbon-design-system/gatsby-theme-carbon/issues/99

PR #129, opened 2019-05-30
chore: update sass build

vpicone:
This dramatically cuts our style bundle and hot reloading times. We don't ship styles we don't intend on using.

This is a breaking change for anyone who was using carbon-components to build their own custom components, they'll need to include the styles for those components in a sass file.


vpicone:
3 min deploy time and way faster hot reloading. Shoulda done this 4 weeks ago, the project would prob be done haha.

PR #146, opened 2019-06-05
fix(page-tabs): set overflow-x to auto

### LINKED ISSUE #145: PageTabs horizontal scrollbar always shows
Note: this is on `v1.0.1` --
latest Chrome

A horizontal scrollbar always shows in the `PageTabs` components:

Screen Shot 2019-06-05 at 3 26 04 PM

This appears to be the cause: https://github.com/carbon-design-system/gatsby-theme-carbon/blob/master/packages/gatsby-theme-carbon/src/components/PageTabs/PageTabs.module.scss#L17

i.e., `overflow-x: scroll` instead of `overflow-x: auto` 🤔

jendowns:
Resolves #145

Prevents horizontal scrollbar (or the "space" for the scrollbar) always showing even when the page tabs aren't overflowing their container.

### DIFF
--- a/packages/gatsby-theme-carbon/src/components/PageTabs/PageTabs.module.scss
+++ b/packages/gatsby-theme-carbon/src/components/PageTabs/PageTabs.module.scss
@@ -14,7 +14,7 @@
   width: 100%;
   list-style: none;
   padding: 0;
-  overflow-x: scroll;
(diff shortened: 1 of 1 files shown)


vpicone:
Woooooo thanks Jen!

PR #160, opened 2019-06-11
feat: extend Video component / image/caption spacing fix

### LINKED ISSUE #15: Video component
- [x] Need to move all styles related to the video component from the `page.scss` to a `video.scss` file. 
- [ ] Custom play button
- [ ] Custom preview image

alisonjoseph:
- Extends the video component to take a local `src` in addition to Vimeo.
- Updates the image spacing to work for gif and svgs 
- Updates caption spacing to work after videos


packages/gatsby-theme-carbon/src/components/Caption/caption.scss
```diff
@@ -2,7 +2,17 @@
   @include carbon--type-style('body-short-01');
   padding-top: $spacing-03;
   margin-bottom: $spacing-06;
-  margin-top: -$spacing-06; // to account for default spacing below images
+}
+
+// Caption after images
+.gatsby-resp-image-wrapper + .#{$prefix}--caption,
```
vpicone: Do we need this if we're targeting the `img` as well?

packages/gatsby-theme-carbon/src/components/Video/video.scss
```diff
@@ -2,6 +2,11 @@
   margin-bottom: $spacing-08;
 }
 
+// video
+.#{$prefix}--video video {
```
vpicone: Do we need the extra element selector here? Don't think we wanna add specificity.

packages/gatsby-theme-carbon/src/components/Video/Video.js
```diff
@@ -4,32 +4,51 @@ import { settings } from 'carbon-components';
...
-            />
-          </div>
-        </div>
-      </div>
-    );
-  }
-}
+  return (
+    <div className={`${prefix}--video`}>
+      {src === '' && vimeoId === '' ? (
```
vpicone: This won't work if they're undefined.

I think we should just do this with proptypes, rather than rendering a paragraph tag maybe?

packages/gatsby-theme-carbon/src/components/Video/Video.js
```diff
@@ -4,32 +4,51 @@ import { settings } from 'carbon-components';
...
+        <p>Please add a vimeoId or src to your video component</p>
+      ) : null}
+      {src !== '' ? <>{videoVid}</> : null}
```
vpicone: {src ? videoVid : vimeoVid}

packages/gatsby-theme-carbon/src/components/Caption/caption.scss
alisonjoseph: Yes, because the actual image tag is nested inside so we need to target this for the sibling selector to work

packages/gatsby-theme-carbon/src/components/Video/video.scss
alisonjoseph: Updated the component to remove the extra container divs

alisonjoseph:
Made some update @vpicone thoughts?

vpicone:
@alisonjoseph looks great, I'm just gonna update the proptypes and maybe do a quick sass module conversion then we'll be good to go

Issue #105, opened 2019-05-24
refactor: convert css-in-js solutions to sass modules


vpicone:
1. Convert existing components

- [x] markdown
- [x] anchor links
- [x] aside
- [x] video
- [ ] left nav
- [ ] main
- [ ] edit
- [ ] container
- [ ] homepage components

2. Distribute sass variables to sass modules globally

PR #170, opened 2019-06-13
Markdown updates

### LINKED ISSUE #68: Typography updates
- Update vertical rhythm to be more consistent and scaleable

### LINKED ISSUE #147: Too much top padding + margin (~128px) if page starts with a heading
Note: this is on `v1.0.1` -- 

If the first element in a page is a heading, the padding-top of the page's grid combined with the margin-top of the heading can add up to 128 pixels --

Screen Shot 2019-06-05 at 3 36 46 PM
Screen Shot 2019-06-05 at 3 37 10 PM

It would be great if the default styling was such that headings at the top of a page had no margin-top value 🤔 

As it stands, this is difficult for us to target & resolve ourselves because of the generated class names, so it would be nice to see it resolved here instead.

### LINKED ISSUE #169: Bring in spacing updates from IDR site
https://github.com/IBM-Design/ai-design/blob/master/src/styles/global.scss#L65

sync w/ Lawrence on specifics if needed
https://www.ibm.com/design/ai/

vpicone:
closes #169, closes #68, closes #147 

begins work on #105 

This PR also separates out the MDXProvider from the components as passes the compmonents in as default props. In this way, devs can shadow the Provider and spread the default components that they don't want to override.

Major markdown components now have their styles colocated `components/markdown/Markdown.module.scss` which allows for an easier whole-sale override.


vpicone:
Since the "h1" role effectively consumed by the page header, we now style H1 tags identically to "h2" tags, except `h1`s now have `margin-top: 0`.

This means the start of your page can be either a `PageDescription` or an `h1` tag (single #). We recommend the former. For most content `PageDescription` → `AnchorLinks` creates a nice hierarchy.

packages/example/src/pages/components/markdown.mdx
```diff
@@ -35,6 +35,12 @@ Combined emphasis with **asterisks and _underscores_**.
 Strikethrough uses two tildes. ~~Scratch this.~~
 ```
 
+## Headers
```
vpicone: Adds documentation around recommended IA and notes the utility of h1s for starting pages.

PR #175, opened 2019-06-17
fix: remove margin-top for first-of-type headings in main grid container

### LINKED ISSUE #147: Too much top padding + margin (~128px) if page starts with a heading

jendowns:
Resolves #147

We were seeing too much padding on some of our pages that start with a heading, specifically an `h2`. So this rule should remove that in the heading is the first direct descendent of the main `bx--grid` class for a page 🤔 

I remove `h1` since `margin-top` is removed elsewhere. I could remove `h3` - `h6` if you want. I just included them just in case 😅

### DIFF
--- a/packages/gatsby-theme-carbon/src/styles/_page.scss
+++ b/packages/gatsby-theme-carbon/src/styles/_page.scss
@@ -29,11 +29,11 @@ body {
     padding-left: 400px;
   }
 
-  > h2:first-of-type,
-  > h3:first-of-type,
-  > h4:first-of-type,
-  > h5:first-of-type,
-  > h6:first-of-type {
+  > h2:first-child,
+  > h3:first-child,
+  > h4:first-child,
+  > h5:first-child,
+  > h6:first-child {
     margin-top: 0;
   }
 }
--- a/packages/gatsby-theme-carbon/src/styles/_page.scss
+++ b/packages/gatsby-theme-carbon/src/styles/_page.scss
@@ -29,7 +29,6 @@ body {
     padding-left: 400px;
   }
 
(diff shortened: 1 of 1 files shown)


vpicone:
@jendowns I believe this removes the margin-top of say an h2 that might come after a PageDescription, the first one (regardless of whether it's at the top of the page or not). Unfortunately the first-child selector doesn't jive with SSR, otherwise we'd just use that I think.

We recently made that change to the h1 tag to try and accommodate headers at the top of the page. I added some documentation around this problem as well let me know if it's not helpful or doesn't solve your use case 

https://gatsby-theme-carbon.now.sh/components/markdown#headers-8

jendowns:
Thanks! Just updated it to `first-child`

Not sure how an SSR issue comes up in *this* case but not others usage cases in this repo or `carbon-website`? 🤔 

About the proposed solution: that's not what we are using in this case & I think it would be nice to avoid being restrictive here.

jendowns:
@vpicone Spoke to the folks our writing content & the `h1` updates are okay with them especially since they are styles the same, as you said elsewhere

So considering that I'll just close this out -- thanks!

PR #181, opened 2019-06-18
fix: add first-child rule and add back h1 styles w/ docs

vpicone:
This reverts the style changes to the `h1` tag from `v1.2.0` in favor of a much more flexible approach to preserving the lead space of content containers. Instead of limiting the starting components, the first child of each page's content container will have it's margin-top removed. 

For most pages, we recommend starting with a `PageDescription` followed by `AnchorLinks` if the content is long enough.

**Note:** The title located in each page header is an `h1`. Some resources suggest you should only have a single `h1` per page, while others indicate it [doesn't matter](https://www.youtube.com/watch?v=WsgrSxCmMbM). As a rule, we recommend using `h2`s as section headings for a consistent hierarchy in your markdown.


packages/example/src/pages/components/markdown.mdx
```diff
@@ -37,12 +37,10 @@ Strikethrough uses two tildes. ~~Scratch this.~~
 
 ## Headers
 
-Since the `h1` role effectively consumed by the page header, we style `h1` tags the same as `h2` tags. The only difference is the lack of a margin-top on `h1` components.
-
-This means the start of your content should be either a [PageDescription](/components/PageDescription) or an `h1` tag (single #). We recommend the former strategy. For most content, we think `PageDescription` → `AnchorLinks` creates a nice hierarchy.
-
 # H1
 
+**Warning:** The title located in each page header is an `h1`. Some resources suggest you should only have a single `h1` per page, while others indicate it [doesn't matter](https://www.youtube.com/watch?v=WsgrSxCmMbM). As a rule, we recommend using `h2`s as section headings for a consistent hierarchy.
```
jendowns: > Some resources suggest you should only have a single `h1` per page ...

Which resources are those?

----

> while others indicate it [doesn't matter] ...

😉 I can see why you included this video & it's a great perspective to have. But perhaps it would be more valuable to consumers  building a sensible hierarchy if you linked to an a11y or web standards resource as well?

jendowns: @vpicone So revisiting this -- this video mentions `h1`s in a page template. So I'm assuming he's implying that it's fine to have `h1`s in new `sections`, or in a `header`, or even in a `nav` (all correct 👍though I wish he would have said as much).

But this is not a complete picture of semantic markup, nor does it provide any context around how proper heading hierarchy helps with a11y / screen readers.

Could you please consider replacing this video with something more comprehensive? The video's lack of context around semantic markup & a11y seems lacking as the sole reference for *this* documentation! 😅 

cc @dakahn -- do you have any thoughts?

vpicone: @jendowns I assumed the google webmasters vid would account for a11y as well as SEO which in hindsight was definitely an oversight. I think having a harder stance against using multiple h1's makes sense from both a design and a11y perspective, definitely a win-win.

I think for extra measure, I'm just going to remove the H1 from the markdown page all together. Thanks again for the input and great resource, I learned a lot.

jendowns: No worries at all!

Issue #180, opened 2019-06-18
"<em>" not rendering italic text


alisonjoseph:
No styles associated with `<em>` 

should be italic

vpicone:
closed by #183

PR #183, opened 2019-06-19
feat: add new global scss file to inject global imports into theme

vpicone:
## New
Developers can now inject imports and other global values and dependencies into the theme style bundle by shadowing `src/gatsby-theme-carbon/styles/_global.scss` in their projects. Shadowing this file allows users to override sass variables and inject global dependencies before the application styles are compiled.  

This **optional** technique is preferred to the previous recommendation of shadowing the `Layout.js` component, particularly when including `node_module` dependencies that assume a single bundle (like `carbon-components`). For your own application styles, you can import a stylesheet into your site's `gatsby-browser.js` or continue shadowing the `Layout` component and importing it there.

#### **TLDR**
* If needed, shadow `_global.scss` to inject global dependencies such as `carbon-components`
* `_global.scss` does nothing if it's not shadowed 
* `_global.scss` gets injected after our globals but _before_ theme styles. 
	
	If you want to override our component styles, you'll need to shadow those style sheets separately.


vpicone:
Closes #180

PR #189, opened 2019-06-20
chore: update list and nested list styles

### LINKED ISSUE #187: Ordered lists need more spacing
If there are more than 9 items in an ordered list we need more spacing

Screen Shot 2019-06-20 at 11 36 56 AM

vpicone:
closes #187, also adds improved support for nested lists.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,11 +16,20 @@ npx gatsby new my-carbon-site https://github.com/carbon-design-system/gatsby-sta
 
 2. **Start developing**
 
-Navigate into your directory and start it up
+Navigate into your directory and start it up using one of the following snippets. You can tell which command to use based on the lock file at the root of your project (`yarn.lock` for yarn and `package-lock.json` for npm).
+
+#### yarn
 
 ```sh
     cd my-carbon-site/
-    gatsby develop
+    yarn develop
+```
+
+#### npm
+
+```sh
+    cd my-carbon-site/
+    npm run develop
 ```
 
 3. **Make some changes!**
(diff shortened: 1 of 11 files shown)


alisonjoseph:
I'm pretty sure ordered list numbers are supposed to be left aligned, which is why we have the issue w/ the overlap. I think it makes more sense to align them how you have it to fix the issue, but it doesn't match up with what Carbon has.

I would sync up w/ design, maybe @jeanservaas on this before we merge this in.

vpicone:
@alisonjoseph Is this an issue somewhere on a live site? If not I'm down for scrapping it in favor of a carbon fix.

jeanservaas:
@vpicone

So this was the original intent of the design -- that the length of the largest number pushes the distance of the list content over accordingly.

I don't think right aligning the numbers in this instance is the way to go in that it really undermines our keyline guidance that type aligns with type etc.

Also, in the examples, the nested ordered items should have letters and the nested un-ordered items should have bullets.

image

vpicone:
I mentioned in person but dynamically controlling the space between the numbers and content based on list length isn't possible. To keep both left aligned numbers and left aligned text, we need to increase the space between both. This would happen even when the list is <10 items which I think is the case the majority of the time.

I put together a code pen to see what that increased space might look/feel like. You can toggle to higher number sets with the buttons: https://codepen.io/vpicone/pen/GbWdpM

vpicone:
Opened https://github.com/carbon-design-system/gatsby-theme-carbon/issues/191

To track the list fix and eventually migrate once it's good.

PR #199, opened 2019-06-26
fix: remove carbon specific homepage styles

alisonjoseph:
- Fixes bug where the grid on the homepage was ignoring the max-width because of a 100% rule we had copied over from Carbon site
- Removes css for dot pattern that is specific to the Carbon site

FYI @sstrubberg


sstrubberg:
Confirmed that this fixed the cards scaling too big.

vpicone:
@alisonjoseph taking away the max-width broke wide screens:

Screen Shot 2019-07-03 at 4 29 51 PM

alisonjoseph:
That's what I get for working on my small 13 inch laptop screen. Looks like we need a way for the background to go 100% but the content to stay within the max-width for the grid.

PR #206, opened 2019-06-28
feat: add accordion components and examples

SilverFox70:
This is the implementation of the Carbon Accordion component(s) used on IBM Brand Center.


packages/example/src/pages/components/Accordion.mdx
```diff
@@ -0,0 +1,98 @@
+---
+title: Accordion
+---
+import { AccordionGroup, Accordion, AccordionItem } from "../../../../gatsby-theme-carbon";
```
vpicone: Yarn workspace ensures we use the component from the theme without a direct import.

```suggestion
import { AccordionGroup, Accordion, AccordionItem } from "gatsby-theme-carbon";
```

packages/gatsby-theme-carbon/src/components/AccordionItem/Accordion.module.scss
```diff
@@ -0,0 +1,85 @@
+:global {
+  @import "~carbon-components/scss/components/accordion/accordion";
+}  
+
+.gtc__title {
```
vpicone: Since the classes are generated programattically you don't have to prefix them.

packages/gatsby-theme-carbon/src/components/AccordionGroup.js
```diff
@@ -0,0 +1,58 @@
+import PropTypes from 'prop-types'
+import React from 'react'
+import styled from '@emotion/styled'
+import { settings } from 'carbon-components'
+
+const { prefix } = settings
+
+const AccordionGroupHeader = styled.h3`
```
vpicone: We want to avoid mixing css-in-js with sass modules

packages/gatsby-theme-carbon/src/components/AccordionItem/AccordionItem.js
```diff
@@ -0,0 +1,152 @@
+/**
```
vpicone: Definitely don't want to be copy and pasting carbon-components-react source code. This will break if we make any changes to the react or carbon components libraries.

Try, instead, targeting the classNames exposed by default from the react library from your own style sheets. You can supply a custom class name to the accordion item, then target the title, content, heading etc from the style sheet.

```
.accordionItem :global(bx--accordion__title) {
	// title styles go here
}
```

packages/gatsby-theme-carbon/src/components/AccordionItem/Accordion.module.scss
SilverFox70: Understood. My goal is to avoid was a) to avoid a name collision with 'title' and b) to make it clear where this class was coming from. I can rename the above and remove the notion of prefixes from the other classes.

packages/gatsby-theme-carbon/src/components/AccordionItem/AccordionItem.js
SilverFox70: Ah, thank you for this1 This was at the root of the question I was trying to ask the other day - how to target the internal classes without explicitly adding them into the react component itself.

packages/gatsby-theme-carbon/src/components/AccordionItem/Accordion.module.scss
SilverFox70: Should be resolved now.

packages/gatsby-theme-carbon/src/components/AccordionGroup.js
SilverFox70: Replace css-in-js with sass module.

packages/gatsby-theme-carbon/src/components/AccordionItem/AccordionItem.js
SilverFox70: Supply custom class name then target elements as needed.

vpicone:
Hey @SilverFox70 you totally nailed it with this PR, but rules like this one and the excessive !importants in the style sheet are concerning to me as I can't easily tell why they're there or for what content they're meant to style.

```
.accordionItem .bx--accordion__content ol li p {
  padding-right: 0rem !important;
}
```

I also don't think the AccordionGroup component is necessary.

I think it'd be best to go with the more bare-bones/flexible option here: https://github.com/carbon-design-system/gatsby-theme-carbon/pull/212/files

After some tweaks that is.

If you guys have unique styles it would be way easier for you to shadow and that style sheet with a one rule than for anyone else to do the opposite.

vpicone:
Closed in favor of #212

Issue #59, opened 2019-05-14
Enhancement[Video Component]: Making video component more robust


jeanservaas:
@shixiedesign commented on [Wed Jan 02 2019](https://github.com/IBM/carbon-elements/issues/216)

- [x] custom play button
- [x] caption support
- [x] need to support mulitple aspect ratios

Dev issue for issue https://github.com/carbon-design-system/carbon-website/issues/473

![image.png](https://images.zenhubusercontent.com/5b96e742f28d730b01a470c0/ac7721c2-df30-45f2-8b1f-2bc52946a72a)

alisonjoseph:
@abbeyhrt this was the video component we were talking about. there will be more to it, but the basic enhancement to start with is to be able to have a custom play button the same as we have here https://www.ibm.com/design/language/elements/2x-grid (on the second video down, the first one is vimeo and doesn't have it)

vpicone:
@alisonjoseph @abbeyhrt is this closed?

alisonjoseph:
yep

PR #215, opened 2019-07-03
chore: add IDL video design to video component

abbeyhrt:
https://gatsby-theme-carbon-qh7nqqryh.now.sh/components/Video/

 **Changed**

- edited the Video component to have an overlay with play/pause button that matches the ibm design website. 
- changed Video.module.scss to create the styles needed to  make the video play/pause button match the original.
- ran `npx carbon-upgrade@canary migrate @carbon/icons-react 10.3.0 10.4.0` so that icon references that used the direct path have been automatically updated to use the module path

Before: 
Screen Shot 2019-07-02 at 3 54 54 PM

What it should match: 

Desired-behavior

What it looks like now: 
Current-behavior

ref #https://github.com/carbon-design-system/gatsby-theme-carbon/issues/59

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -2,7 +2,7 @@
   "packages": [
     "packages/*"
   ],
-  "version": "1.3.3",
+  "version": "1.3.4",
   "npmClient": "yarn",
   "useWorkspaces": true,
   "publish": {
--- a/lerna.json
+++ b/lerna.json
@@ -2,7 +2,7 @@
   "packages": [
     "packages/*"
   ],
-  "version": "1.3.4",
+  "version": "1.3.5",
   "npmClient": "yarn",
   "useWorkspaces": true,
   "publish": {
--- a/packages/example/package.json
+++ b/packages/example/package.json
@@ -1,7 +1,7 @@
 {
   "name": "example",
   "private": true,
-  "version": "1.3.3",
+  "version": "1.3.4",
   "license": "Apache 2.0",
   "scripts": {
     "develop": "gatsby develop",
@@ -13,7 +13,7 @@
   },
   "dependencies": {
     "gatsby": "^2.4.2",
-    "gatsby-theme-carbon": "^1.3.3",
+    "gatsby-theme-carbon": "^1.3.4",
     "react": "^16.8.6",
     "react-dom": "^16.8.6"
   }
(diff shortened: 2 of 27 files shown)


abbeyhrt:
@vpicone I'm not sure why the deployment failed, seems to do with the gatsby-theme-carbon update when I merged master in. Do you know what's going on? Any help would be greatly appreciated!

abbeyhrt:
@dakahn here's the link to the specific video, it's the second one down! https://gatsby-theme-carbon-da2itj8xr.now.sh/components/Video 
Thank you!

abbeyhrt:
Added style for the focused state and seems like it's all good on accessibility for now!

abbeyhrt:
Added the motion!

dakahn:
Testing on Windows 10 in latest Chrome and latest Firefox, with NVDA 2019

I see this got merged 🎉. Just wanted to leave some a11y notes from my testing for future work

I see that the video is without expected controls by default (full screen, closed caption disable/enable, scrubbing, and volume)  -- is that always the case? _I don't have much experience with our video player, it's stated goals or what user needs it solves._ It's also strange that these controls are can be accessed, but only by mouse users via a right click. 😕 

**Problem - focus state contrast**
The focus indicator for the play/pause button needs work. Against the edges of the button we have a contrast of 1.16:1 ignoring the button color itself (which we shouldn't do) the focus outline at various scenes in the video is hovering around 2:1 contrast ratio territory.

If we're committed to the big hovering button floating independent of a toolbar we need a stronger focus state. Because the foreground focus state must take the background color into account (and the way we've built our player that background color is a moving video) the best solution I can fathom is an inversion of colors on the button itself -- the goal is 3:1 or better for components against their backgrounds 👍 

That said, this is currently a problem also shared by the worlds most well worn video player Youtube
2019-07-08 16_22_47-Window

Looks fine against a dark enough background, but sky diving videos are gonna get dinged by WCAG 2.1 for sure

**Problem - poor screen reader support**
The video's state isn't announced by NVDA. By default the video is paused and the only tab stop in the player is the play button. When that button is pressed no announcement is made (nothing at all is said), and the user doesn't really have any idea what's playing -- or that the play button they were focusing before has now mighty morphed into a pause button. The only way they'd know that is to tab away and then back to the button.

For solutions here's a guide[ on creating accessible video player controls from MDN](https://developer.mozilla.org/en-US/docs/Learn/Accessibility/Multimedia#Accessible_audio_and_video_controls)

Though more thorough browsers/screen reader testing should be done ♿️ ☮️

Issue #211, opened 2019-07-02
Add mobile icons to the theme


sstrubberg:
https://www.gatsbyjs.org/packages/gatsby-plugin-manifest/#hybrid-mode-configuration

[Touch icons.zip](https://github.com/carbon-design-system/gatsby-theme-carbon/files/3352175/Touch.icons.zip)

tay1orjones:
Tangentially related - This week I ran into an issue with this plugin. It doesn't allow you to configure separate artwork for different sizes. Our ideal set up is a transparent favicon for 16px-64px, and then separate artwork with a background for larger sizes, including `apple-touch-icons`.

Additionally we wanted separate artwork for dev/prod so tabs are easily distinguishable when you have them open side by side. We resorted to manual import and config of these in the Layout via Helmet.

Will we have to shadow the entire gatsby-config to overwrite the settings of this plugin, or can we shadow only the plugins we need to reconfigure? To support the dev/prod case, will we need to shadow the entire layout as well?

vpicone:
@tay1orjones you can shadow individual plugins, just like I show in the readme.

It strikes me as odd that you'd want to bring in the overhead of separate art between dev and prod just for the convenience of distinguishing them. Generally you want to keep your dev and prod environments as close as possible.

However, you can definitely shadow the Layout but keep in mind, you aren't meant to re-implement the Layout logic. It could be as simple as.

```jsx
import Layout from 'gatsby-theme-carbon/blah blah'

const CustomLayout = (props) => {
	return (
		<>
			<Helmet />
			<Layout {...props}>
		</>
	)
}
```

vpicone:
@sstrubberg Can you elaborate on what mobile icons are and what you need the theme to do here?

sstrubberg:
![IMG_378E9082537F-1](https://user-images.githubusercontent.com/15822070/60615011-c60ced00-9d93-11e9-9e89-9221257c6b6f.jpeg)

The icon on the upper-left is an example of what happens when you save w3.ibm.com/design as an icon on your home screen. The icon in the lower-right is what happens when i do the same thing for IDR using the Gatsby theme.

Ideally we could get the rebus setup on the template.

vpicone:
Turns out apple just automatically applies a black background so our bee just blended in. It was there, just hidden.

PR #218, opened 2019-07-08
fix: allow favicon override

vpicone:
closes https://github.com/carbon-design-system/gatsby-theme-carbon/issues/211

Allows for local path to override default favicon


vpicone:
closes https://github.com/carbon-design-system/gatsby-theme-carbon/issues/211

PR #238, opened 2019-07-11
Refactor: update vimeo player

abbeyhrt:
*Changed* 

- Updated vimeo player to match the visual of videos on the IDL site. 
Since I can't find a way to edit the actual vimeo videos, there is just a button overlay on top of the vimeo button and the vimeo button is still visible at times. I don't currently know a way around this 
Before: 
Before

Now: 
Now

PR deploy site: https://gatsby-theme-carbon-git-fork-abbeyhrt-chore-update-vimeo-player.carbon-design-system.now.sh/components/Video


abbeyhrt:
With all this new work on the video and vimeo players, should we consider separating them into different components? Like `<Video />` and `<Vimeo />` or something?

Issue #162, opened 2019-06-12
"Edit on Github" link


tay1orjones:
Would be nice to have the `Edit on GitHub` link present at the bottom of each page, like carbondesignsystem.com

@vpicone what was the alternative idea you had besides reading the git repo from package.json to build the link?

vpicone:
closed with #239

PR #239, opened 2019-07-11
feat(editlink): add to default template, style updates, associated docs

tay1orjones:
Just took a quick shot here, let me know your feedback and I can revise as necessary

- Migrated the previous emotion styles to the scss module
- Added the component to the default template
- Updated config to accept params for `baseUrl` and `subDirectory`
- Added documentation to the example site


tay1orjones:
Closes #162

vpicone:
Awesome!!! I'll dive more into this tomorrow, but looks solid at first glance.

packages/gatsby-theme-carbon/src/components/EditLink/EditLink.module.scss
```diff
@@ -1,6 +1,11 @@
 .link {
-  color: $text02;
+  color: $text-02;
```
vpicone:
```suggestion
  @include carbon--type-style('body-short-01');
  transition: color $transition--base;
  color: $text-02;
```

packages/gatsby-theme-carbon/src/components/EditLink/EditLink.module.scss
```diff
@@ -1,6 +1,11 @@
...
   &:hover {
-    color: $text01;
+    color: $text-01;
   }
 }
+
+.row {
+  position: relative;
+  bottom: calc(-160px + 0.875rem + 32px);
```
vpicone:
```suggestion
  top: 9rem;
```
We want `calc(160px - 14px) = 144px = 9rem`. Spec is from [this comment](https://github.com/carbon-design-system/carbon-website/issues/862#issuecomment-467622946)

tay1orjones:
@vpicone updated with requested changes

vpicone:
Great work. You nailed it!

tay1orjones:
@vpicone great point, thanks for the catch. I just pushed an update with that change.

vpicone:
@tay1orjones aw dang just realized this doesn't work if the site is an `index.mdx` inside of a directory (like the demo page). I think the solution is a page query that shares the file name. Unfortunately page queries are kind of a PITA for gatsby-mdx. We might need to do this in `gatsby-node`.

vpicone:
For the `index.mdx` fix, I added a hook to `gatsby-node` so we can provide templates with additional page context (like like the page's file path).

```js
// We need to provide the actual file that created a specific page to append links for EditLink.
// We can't do page queries from MDX templates, so we'll add the page's relative path to context after it's created.
// The context object **is** supplied to MDX templates through the pageContext prop.
exports.onCreatePage = ({ page, actions }) => {
  const { createPage, deletePage } = actions;
  const [relativePagePath] = page.componentPath.split('src/pages').splice('-1');
  deletePage(page);
  createPage({
    ...page,
    context: {
      ...page.context,
      relativePagePath,
    },
  });
};

```

packages/gatsby-theme-carbon/src/components/EditLink/EditLink.js
```diff
@@ -1,40 +1,48 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import { css } from '@emotion/core';
+import { graphql, useStaticQuery } from 'gatsby';
 
-import { link } from './EditLink.module.scss';
+import { link, row } from './EditLink.module.scss';
 
-const repository = 'https://github.com/temporary/master/tree';
+const EditLink = ({ relativePagePath, repository: repositoryProp }) => {
```
vpicone: Moved this to it's own query. Ideally data fetching would be separate from the component itself but we definitely don't want them sharing queries in case one gets shadowed.

tay1orjones:
Looks good to me, thanks!

vpicone:
@alisonjoseph yeah it doesn't render unless a url is provided as a theme option.

PR #242, opened 2019-07-12
Code snippet

vpicone:
This adds syntax highlighting and better copy button integration to the code snippet component. 

In addition you can now provide a title (file path) to the code snippets. If a URL is provided for the source, we'll instead generate a link that points to the provided url 🎉


vpicone:
@alisonjoseph i talked with @jeanservaas about it yesterday. The whole reason for that code text is to balance out the space you’re talking about, and we both hate it haha. You can also just provide a regular string and it’ll be a title that fills that space. I played with having it default to “Code:” but it feels redundant.

One alternative that I think looks better is to have the copy button be an actual button rather than blend in with the background. That way it doesn’t look as terrible if it hovers over the content. Another might be a sort of vertical toolbar that goes up and down the right side.

vpicone:
Neither of the types of code snippets on Carbon had line numbers so I left them off for now.

alisonjoseph:
I agree on leaving the word "Code:" out, but still think it looks super weird without anything up there. 


What about adding the line, with the copy button above?
Screen Shot 2019-07-12 at 8 35 42 AM

looks a little less odd w/ out a title or url
Screen Shot 2019-07-12 at 8 36 28 AM

or could default to this w/ out the title or url, which matches the core Carbon component currently
Screen Shot 2019-07-12 at 8 37 33 AM

vpicone:
Yeah I think that’s probably best, the only issue is sometimes the code goes under the button and creates a weird mask/overlay. We could do a radial gradient or shadow to make it look like the code fades out just under the button or make the button inverted at smaller media queries. If it’s not the same color as the background it makes sense to be above the text.

abbeyhrt:
This is great! It looks so much better with the syntax highlighting. One question though, do we care that it's highlighting all keywords? Like in this code snippet for example, from and in are highlighted. 
Screen Shot 2019-07-12 at 9 49 08 AM

vpicone:
@jeanservaas @laurenmrice you can see each of the different design permutations [here](https://gatsby-theme-carbon-git-fork-vpicone-code-snippet.carbon-design-system.now.sh/components/code-snippet).

@alisonjoseph and I decided we'd like to try not including the "show more" functionality to start. We can add it later on if needed, but the version without the show more button on the design website seems more useful by default.

vpicone:
@abbeyhrt good catch! I think I was a little overzealous when converting the theme to `react-prism-renderer`.

PR #244, opened 2019-07-12
chore: update video spacing to match IDL site

### LINKED ISSUE #226: Update video spacing
Spacing below video is too tight, see IDL Philosophy page for correct spacing

https://www.ibm.com/design/language/philosophy

abbeyhrt:
Closes #226 

*Changed*

- updated margin-bottom token form $spacing-08 to $carbon--layout-05, changing the px below the video from 40 to 64 to be inline with the IDL site. 

Before: 
Screen Shot 2019-07-12 at 10 30 07 AM

After: 
Screen Shot 2019-07-12 at 10 30 18 AM


packages/gatsby-theme-carbon/src/components/Video/Video.module.scss
```diff
@@ -5,7 +5,7 @@
 }
 
 .video-container {
-  margin-bottom: $spacing-08;
+  margin-bottom: $carbon--layout-05;
```
vpicone: A few months ago we added all the layout values to the spacing scale so we wouldn't have to alternate between the two. You also don't need the carbon prefix.

```suggestion
  margin-bottom: $spacing-10;
```

alisonjoseph:
I haven't seen any requests from @jeanservaas or others to adjust the Tabs or DoDont spacing, we should probably sit down at some point and schedule a spacing audit of everything with design though as there are multiple issues around spacing in Mike's box note.

abbeyhrt:
@vpicone committed the suggestion you made and I think this should be good to go then!

PR #246, opened 2019-07-12
chore: update anchor link spacing

### LINKED ISSUE #225: Update anchor link spacing
Spacing below anchor links is too small, see IDL site for correct spacing.

https://www.ibm.com/design/language/philosophy

abbeyhrt:
Closes #225 

*Changed* 

- Updates `margin-bottom: $spacing-02;` -> `margin-bottom: $spacing-03;`  changing the sapce under AnchorLinks from 4px to 8px, to match the IDL site. 

Before: 

Screen Shot 2019-07-12 at 12 20 04 PM

Now: 
Screen Shot 2019-07-12 at 12 16 49 PM

Matching this spacing: 
Screen Shot 2019-07-12 at 12 16 38 PM


alisonjoseph:
I'll defer to @jeanservaas on this, but I should have been more clear in the original issue. Sorry! 😨  I think the issue was the spacing below the block of links, but it appears the spacing between might not have been quite right either.

Screen Shot 2019-07-12 at 1 45 49 PM

abbeyhrt:
That's my bad! I'll add the spacing below the group of anchorlinks in real quick, do you want me to take out the added space between links?

abbeyhrt:
I added the spacing so that it matches the IDL site at different viewports, kept in the spacing between Anchorlinks for now

jeanservaas:
Hey @abbeyhrt 

Don't think we need spacing between anchor links... I think on both IDL and Carbon those guys are using $expressive-heading-03 and just getting the responsive line height from that. We def. do want to match what IDL is doing, but I think that's what they're doing.

Also...hmmm. Not sure exactly what I'm reviewing here because the margin bottom on the anchor links 
usually just depends on what comes after them right? I can't find an instance on Carbon where a video follows them... they're usually followed by an H2 ... so in that case there's 80px of distance.

I do see this example in IDL... but correct me if I'm wrong @vpicone but isn't this part of the larger vertical spacing issue that we've been talking about? 

image

Sorry for the confusion @abbeyhrt -- I'll hit you up tomorrow and we can sort this out!

vpicone:
@jeanservaas yeah for the inter-link spacing I went with Carbon (4px vs IDL 8px).

@abbeyhrt we can try setting the margin-top on any element following an Anchor list

To do this, you might consider using the adjacent sibling selector and the wild card selector to apply margin tops of spacing-11 to the normal list and spacing-10 to the small lists (the margin tops of h2s and h3s respectively)

alisonjoseph:
@jeanservaas this request came from this box note from Mike as a FYI https://ibm.ent.box.com/notes/487755944342

abbeyhrt:
@vpicone i reverted the other changes I made back to their original state and added in adjacent sibling wild card selectors, let me know if there are any other changes that need to be done!

PR #247, opened 2019-07-12
chore: update to Carbon v10.4 + Other stuff

vpicone:
Main stuff:
* Updates to Carbon v10.4 and CCR v7.4 and fixes 10.3/4 shell problems
* Fixes issues with Carbon Tabs component from 7.3

Other:
* Address styling issues around Carbon lists
* Adds a fix/workaround for the gatsby-remark-images blur-up and padding issue
* Add tab architecture guidelines


vpicone:
Still need to check on lists again.

packages/example/src/pages/guides/navigation/tabs.mdx
```diff
@@ -0,0 +1,54 @@
+---
+title: Tabs
+tabs: ['Sidebar', 'Tabs']
+---
+
+<PageDescription>
+
+You have the option of adding tabs to your pages. This is really only recommended if you have many pages with the same tabs. If you just have one page a set of tabs, it might be better for discoverability to make those menu items.
```
abbeyhrt: Just a suggestion to add clarity to this paragraph, totally don't have to do it!

```suggestion
You have the option of adding tabs to your pages. This is only recommended if you have several pages that will use the same tabs. If you only have one page with a set of tabs, it might be better for discoverability to change the tabs to menu items.
```

packages/example/src/pages/guides/navigation/tabs.mdx
```diff
@@ -0,0 +1,54 @@
...
+      path: /guides/shadowing
+    - title: Styling
+      path: /guides/styling
+    - title: Navigation
+      path: /guides/navigation/sidebar
+```
+
+## File structure
+
+Lets check out the directory structure for this page. Notice how you'll add another directory that corresponds with the one in the YAML file.
```
abbeyhrt: This may be way too specific of me,  my bad if it is!
```suggestion
Let's check out the directory structure for this page. Notice how you'll add another directory that corresponds with the one in the YAML file.
```

packages/example/src/pages/guides/navigation/tabs.mdx
vpicone: >totally don't have to do it!

You're herby no longer required to tack this on to your suggestions. 

Thank you for the corrections!!

packages/example/src/pages/guides/navigation/tabs.mdx
vpicone: Not at all!

PR #257, opened 2019-07-18
Add stylelint and precommit linting for scss

vpicone:
Adding stylelint already helped catch some duplicate selectors and rules.

The biggest help though is making sure that each stylesheet decreases in specificity. Especially for pseudo selectors/elements.

Docs added in #267


package.json
```diff
@@ -39,6 +40,10 @@
     "lerna": "^3.13.4",
     "lint-staged": "^9.2.0",
     "prettier": "^1.17.0",
+    "stylelint": "^10.1.0",
+    "stylelint-config-css-modules": "^1.4.0",
```
vpicone: gotta take this one out, not needed

PR #259, opened 2019-07-18
chore: update carbon elements to July release

vpicone:


packages/gatsby-theme-carbon/package.json
```diff
@@ -20,21 +20,21 @@
...
-    "@carbon/layout": "^10.2.0",
-    "@carbon/themes": "^10.3.0",
-    "@carbon/type": "^10.2.0",
+    "@carbon/colors": "^10.3.0",
+    "@carbon/elements": "^10.4.0",
+    "@carbon/grid": "^10.4.0",
+    "@carbon/icons-react": "^10.4.1",
+    "@carbon/layout": "^10.3.0",
+    "@carbon/themes": "^10.4.0",
+    "@carbon/type": "^10.3.0",
```
alisonjoseph: If we have `@carbon/elements` can we get rid of layout, type, and colors? I believe it includes all of those (but not icons-react)

vpicone: yah good point

alisonjoseph:
That was supposed to be a 👍 ... not the Thailand flag lol

PR #268, opened 2019-07-19
fix: banner width

vpicone:


tay1orjones:
👍 This is the only thing holding up our PR to refactor the Watson IoT design/addons site to use the theme. Could you do a minor/patch release for us when this gets in?

vpicone:
@abbeyhrt we only use the css-in-js here so we can do the dynamic position 'top center' etc. Unfortunately, the way we were wrapping the Grid component caused the Grid's styles to override the module styles. This is a temporary fix until we evaluate how best to use css-in-js for minor things like this.

alisonjoseph:
LGTM! was there an issue associated with this we need to close?

vpicone:
@alisonjoseph negative, I opened https://github.com/carbon-design-system/gatsby-theme-carbon/issues/269 to explore not using emotion for this sort of behavior.

Issue #252, opened 2019-07-16
Spacing/vertical rhythm refactor


alisonjoseph:
We have multiple bugs/issues around spacing, need to sync up w/ @jeanservaas and @sadek to make sure spacing above/below components and headings are correct. Need to look at the site/components as a whole and not just individually.

alisonjoseph:
We have a meeting planned for Tuesday 7/30 to solve this. 

@abbeyhrt would it be possible for you to create a demo page in the theme with al(as many as possible) variations on components before/after headings, paragraphs etc prior to this meeting? We don't need to merge it in, we can keep it in a draft PR / branch, but just somewhere we can work from and see things.

vpicone:
Adding #271 to the list of review needs. Spacing in and among lists needs to be better defined when they contain things like code snippets or images.

abbeyhrt:
@alisonjoseph Sorry i didn't see this earlier! But yes, I can get that done ASAP and definitely before Tuesday 7/30

vpicone:
Adding this here for discussion: https://csswizardry.com/2012/06/single-direction-margin-declarations/

Definitely worth the read before we discuss the sort of design direction would be most scaleable

vpicone:
Another great resource: https://every-layout.dev/layouts/stack/

rcwakelin:
@alisonjoseph @vpicone Is this closed?

rcwakelin:
Adding Matt @mattrosno to chain

vpicone:
closed by #373

PR #271, opened 2019-07-22
Update index.mdx

connor-leech:
The way that markdown is rendering these ordered lists is all wonky. It thinks each numbered item below "First steps" is its own `<ol>`.

Either all of it or none of it can be bold, so I erred on all of it bold.


vpicone:
Will discuss spacing in and around lists with #252. I included a couple tweaks to hack around MDX's paragraph/list business.

PR #273, opened 2019-07-23
fix: update icon sizing for resource cards

alisonjoseph:
ref https://github.com/carbon-design-system/design-language-website/issues/99

Updates the icon size to be a max 64px width and height for the resource cards. We were previously assuming 32x32 icons, but that is not the case everywhere.


alisonjoseph:
Oh good point. I added a min-wdith back in, I'll just have to do a specific override for the IDL site, they have an icon that's 21x52px wide on this page https://design-language-website.netlify.com/elements/logos/rebus so its displaying too large. 😓

vpicone:
Hmmmm still not working for me

alisonjoseph:
me either 😓 -- I'll look at it tomorrow sometime

vpicone:
Are we good to go here? Did you wanna test it on IDL first?

alisonjoseph:
Just tested on IDL and seems I need to remove the min-width. See any issues w/ that? I didn't on IDL, but might be missing something.

vpicone:
Removing min-width didn't work for me. Adding `position: static` to the gatsby responsive image wrapper did though.

alisonjoseph:
@vpicone ok I think its really working this time

PR #280, opened 2019-07-29
fix: remove janky :active class from tabs component

id0Sch:
Hi,
I looked around in the example tabs page (https://gatsby-theme-carbon.now.sh/components/Tabs/) and noticed that when the mouse is pressed (:active), the tab label acts weird IMO.

ScreenRecording2019-07-29at9

IMO this janky behavior is a bug, and by looking at the carbon design system's storybook(http://react.carbondesignsystem.com/?path=/story/tabs--default) it seems that it is not intended.

If you think this is the intended behavior, feel free to ignore and close this PR :) 
great job on the theme!! it looks and feels awesome!


vpicone:
LGTM Ty!

PR #292, opened 2019-07-31
feat: pass MDX frontmatter title into page `<title>`

emyarod:
this PR passes the MDX frontmatter title into the `<Meta>` component HTML title element. not sure how configurable we want this to be, but this is an option for getting dynamic page titles in the format we're looking for in the IDL site

### DIFF
--- a/packages/example/src/pages/guides/configuration.mdx
+++ b/packages/example/src/pages/guides/configuration.mdx
@@ -148,6 +148,7 @@ To add a link to the bottom of each page that points to the current page source 
 
 - `additionalFontWeights` – add support for additional Plex font weights. Don't forget to specify italics for the additional weights if needed.
 - `mdxExtensions` – change the file extensions processed by `gatsby-mdx` (default ['.mdx', '.md'])
+- `titleType` – the format of the title in your head. `site` for your site title, `page` for your individual page title, or `append` for "title – pageTitle" (default 'page')
 
 ```js
 plugins: [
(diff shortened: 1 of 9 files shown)


vpicone:
@emyarod this looks great thanks. Sorry I was experimenting with the `www` site and might have caused the now deployment issue. Should be fixed with your next commit

Could you add optional description/keywords as well? I know brand center was asking for this.

emyarod:
@vpicone just double checking, would optional description/keywords also be from MDX frontmatter?

vpicone:
@emyarod yeah, that'd be ideal

vpicone:
If they don't specify in frontmatter, it needs to be undefined so as to not override the one from siteMetadata

emyarod:
local build was taking a while so I haven't had a chance to preview this but I wrote a few statements about the frontmatter. let me know if I should add more info @vpicone

packages/example/src/pages/guides/configuration.mdx
```diff
@@ -148,14 +148,16 @@ To add a link to the bottom of each page that points to the current page source
 
 - `additionalFontWeights` – add support for additional Plex font weights. Don't forget to specify italics for the additional weights if needed.
 - `mdxExtensions` – change the file extensions processed by `gatsby-mdx` (default ['.mdx', '.md'])
+- `titleType` – the format of the title in your head. `site` for your site title, `page` for your individual page title, or `append` for "title – pageTitle" (default 'page')
```
alisonjoseph: I think we need some examples here to make it easier to understand what each one means.

vpicone: @alisonjoseph  added

emyarod:
@vpicone yeah looks good to me. wasn't sure what kind of flexibility we were gonna offer WRT page title but this gives users more options. what's causing the lockfile changes though?

vpicone:
@emyarod mdx had a patch bump while I was working on it, nothin major

emyarod:
@vpicone that's fine was just wondering what prompted the change

PR #294, opened 2019-07-31
feat: add ArtDirection component

vpicone:


alisonjoseph:
Not sure about the name "ArtDirection", but otherwise love the simplicity here

vpicone:
@emyarod Yeah, otherwise you might end up duplicating the middle one to hit two media queries or something silly like that.

alisonjoseph:
If someone puts only two images the first will show at mobile, and the second at tablet and above, correct?

vpicone:
@alisonjoseph art direction is the [actual name](https://developer.mozilla.org/en-US/docs/Learn/HTML/Multimedia_and_embedding/Responsive_images#Art_direction) for the thing we're doing here.

I considered more vague names (e.g. Image, Picture, ImageComponent, Images, ResponsiveImages), but they all imply that there is something _other_ than art direction happening.

vpicone:
@alisonjoseph yeah that's correct. I updated the documentation to explain more clearly:

>You can place up to three images inside of the ArtDirection component. The first will be used for mobile, the second for tablet, and the third for desktop. If only two images are provided, the second image will be used for both tablet and desktop.

PR #309, opened 2019-08-05
fix(navigation): explicitly match paths to links for selected state

### LINKED ISSUE #265: Left Nav duplicate selection when similar page titles
When two left nav/pages in a section have similar titles, like`/components/list-builder` and `/components/list`, the nav applies selected state to more than one item at a time.

When `list-builder` is selected, the nav improperly highlights both `list` and `list-builder`

Screen Shot 2019-07-19 at 9 05 42 AM

The nav behaves as expected when `list` is selected

Screen Shot 2019-07-19 at 9 08 35 AM

tay1orjones:
Took a shot at a fix to resolve #265 

Slightly modifies the left nav item to more explicitly check the paths against one another, rather than relying on `includes()`.

### DIFF
--- a/packages/example/gatsby-config.js
+++ b/packages/example/gatsby-config.js
@@ -20,7 +20,6 @@ module.exports = {
     {
       resolve: 'gatsby-theme-carbon',
       options: {
-        additionalFontWeights: ['200', '200i'],
         repository: {
           baseUrl:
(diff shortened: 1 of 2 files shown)


vpicone:
if the fix was this easy i'll be so mad lol. Ty for taking a stab I'll check it out

vpicone:
One thing that looks like it might be an issue from the start is not accounting for path prefixes.

PR #329, opened 2019-08-12
feat: IBM Plex Sans Variable

### LINKED ISSUE #330: Explore using IBM Plex Sans Variable
We've had the Plex variable font in our arsenal for a bit, I think this is a great opportunity to take advantage of it.

Using a single variable font decreases the number of font file requests and the total payload of the font files (particularly on editorial content where there's a high likelihood of more than one font weight). Using the Plex variable also allows us to experiment with more expressive type tokens, potentially taking advantage of the other axis. 

More info on variable fonts: https://youtu.be/0AsAjHPupKQ

vpicone:
closes #330 

This switches our Plex loading strategy to use local variable fonts. When subsetting the variable fonts I also included those pesky arrows.

We cut the requests from 7→2 and the payload from 126kb to 56kb. This also opens the door for experimenting with more expressive type tokens and better interpolation between those tokens.

tenor


shixiedesign:
@vpicone to review this, do we make sure visually the fonts are not different from current deployment? Coz I noticed a few small things between this deployment and `https://gatsby-theme-carbon.netlify.com/` and `https://www.ibm.com/design/language/elements`. Lemme know if I'm reviewing for the correct thing and I will post detailed notes.

vpicone:
@shixiedesign the url for the master branch/most up to date version of the theme is https://gatsby-theme-carbon.now.sh (not netlify).

This doesn't change anything about the font styles, just the Plex file we're using.

shixiedesign:
I noticed these things: 
1. Some types are squashed vertically. 
2. The single quote is no longer square. 
3. `—` and type after the `—` might have gotten overall a bit smaller...not sure if this is actually a bug. 

Aug-12-2019 11-09-24

vpicone:
@shixiedesign woah! will have to investigate more closely... this PR uses the plex files directly (as opposed to google fonts). I wonder if google fonts has been doing something unusual with our font files. Thanks so much for giving this a careful look!

vpicone:
@shixiedesign it looks like I was too restrictive with the characters I took out. Could you do that comparison again now?

PR #332, opened 2019-08-13
refactor: remove addons-website dependency

alisonjoseph:
Brings in BackToTopBtn component, grid additions and shell overrides to this repo. This removes our dependency on the deprecated @carbon/addons-website package. 

### Testing
- Make sure UI Shell looks correct (side nav/header)
- Make sure back to top button looks correct
- Make sure grid component gutter classes are being applied correctly


alisonjoseph:
I actually removed some of the side nav code too 😭  was able to completely get rid of the two header files though, and just add the couple styles we actually needed to the modules file. 

I think we do need that grid code, but should probably PR it to the grid package at some point.

PR #339, opened 2019-08-15
feat: update switcher at lg+ sizes

### LINKED ISSUE #57: enhancement [app switcher]: truncate app switcher
## enhancement [app switcher]: truncate app switcher

**Why:**
The app switcher now is mirroring the left nav behavior almost exactly which is confusing to the user; by keeping the left nav anchored and floating the global app switcher we create a clearer hierarchy between the two.  

<br>

**Where:**
This floating app switcher remains a global element across all practices... we just want to make that function clearer. It does not change color when the local content changes, it operates on a separate logic / independent of the local left nav. 

- [x] All type and core / interative color tokens, dropshadow and most geometries remain the same

- [x] Add subheads

- [ ] shorten app switcher in Max, XL, Lrg and tablet breakpoints (truncate app switcher 24px after the last link item block

- [ ] 1px, Gray 80 stroke animates up 90 degrees to include the "close" icon in the component when the floating app switcher opens

- [ ] In desktop, panel should animate on like an overflow menu; in mobile, the animation remains a side push and still takes over the screen vertically

image

### Design specs:
https://github.com/carbon-design-system/carbon-website/issues/1200

vpicone:
closes #57 

- [x]  shorten app switcher in Max, XL, Lrg and tablet breakpoints (truncate app switcher 24px after the last link item block
- [x] 1px, Gray 80 stroke animates up 90 degrees to include the "close" icon in the component when the floating app switcher opens **(see note about animation)**
- [x] In desktop, panel should animate on like an overflow menu; in mobile, the animation remains a side push and still takes over the screen vertically

animation stuff: Overflow menus have no animation, so to avoid animating the border without animating the expansion I left the border-left animation out. It's also really hard to tell the animation is happening since it would happen while the button is hovered:

Screen Recording 2019-08-15 at 11 56 17 AM


packages/gatsby-theme-carbon/src/components/Header/Header.module.scss
```diff
@@ -42,3 +42,8 @@ header.header {
 .header :global(.bx--header__name) span {
   font-weight: 600;
 }
+
+.header .switcher-button--open {
+  border-left: 1px solid $inverse-02;
+  border-bottom: 1px solid #171717;
```
alisonjoseph: I think this is gray-100 or `$ui-05`

jeanservaas:
@shixiedesign will look at this and suggest a motion treatment for it.

shixiedesign:
@vpicone Hey here's the motion spec for app switcher: 

**Open:** 
`productive, entrance` or cubic-bezier(0, 0, 0.38, 0.9)
duration: `$duration--moderate-01` or 150ms

**Close:** 
`productive, standard`, or cubic-bezier(0.2, 0, 0.38, 0.9)
duration: `duration--fast-02` or 110ms

app-switcher

vpicone:
@shixiedesign animations added

shixiedesign:
@vpicone Nice getting around to it so fast! Two things: 

- Opening/closing the app switcher should change the height of the container, cropping content inside, not squeezing content inside. 
- Closing animation seems to skip the start, and container jumps from full height to, like 10% height. 

**Current:**
Aug-16-2019 09-56-58

**Should be**
Aug-16-2019 10-02-36

vpicone:
@shixiedesign darn you and your slomo! You caught my hack, I’ll take a look at it haha.

PR #340, opened 2019-08-15
Search stuff

vpicone:
Call search from input, use layout effect to ensure dom node is present


packages/gatsby-theme-carbon/src/components/Header/Header.module.scss
```diff
@@ -42,3 +42,8 @@ header.header {
 .header :global(.bx--header__name) span {
   font-weight: 600;
 }
+
+.header .switcher-button--open {
+  border-left: 1px solid $inverse-02;
+  border-bottom: 1px solid #171717;
```
alisonjoseph:
```suggestion
  border-bottom: 1px solid $carbon--gray-100;
```

PR #350, opened 2019-08-16
fix: spacing updates / fixes 

### LINKED ISSUE #359: Body text width: some areas where it seems not to span usual width in Ai site. Not sure if Gatsby of specific to Ai site
https://www.ibm.com/design/ai/ethics/explainability

Looks like the text has room to be longer but something preventing it from going wider.

Screen Shot 2019-08-18 at 2 43 17 PM

alisonjoseph:
Met w/ @jeanservaas to go over spacing for paragraphs, quotes and asides inside of columns. 

closes #359


packages/gatsby-theme-carbon/src/components/Aside/Aside.module.scss
```diff
@@ -3,6 +3,7 @@
   position: relative;
   padding-top: $spacing-04;
   top: -$spacing-04;
+  width: 84%;
```
vpicone: I don’t really remember where this value came from,  can we handle this with padding so we don’t need to target individual elements (like the p’s below). Sometimes I think it’s helpful to assume that the component isn’t being used with mdx. Can we guarantee the internal content will be a paragraph?

alisonjoseph: This value came from Jeanne when sitting down next to me, to get design correct. I can move it to padding.

alisonjoseph:
docs updated

PR #351, opened 2019-08-16
fix: list spacing

### LINKED ISSUE #341: bug(list): Unordered lists spaced incorrectly
Spacing between individual unordered list items is too substantial.

image

The spacing between list items on the current site build is correct:

image

cc: @jeanservaas

alisonjoseph:
closes #341


vpicone:
@alisonjoseph this might also be a result of us re-importing the list styles so let's see if removing the double import fixes it.

alisonjoseph:
@vpicone we still need to remove the margin-bottom, but we should be able to remove the global style

PR #352, opened 2019-08-16
fix: link font size

### LINKED ISSUE #334: bug(Link): Link font size is incorrect
An embedded link is too small:

image

Font size is 0.875rem, while body copy is 1.5rem.

alisonjoseph:
closes #334


vpicone:
I think this might actually be due to us re-importing the styles into www and this component not properly using the _global stylesheet. I'm gonna update it to use the correct method and see if that works first.

vpicone:
I believe this will be fixed once we don't duplicate the link import in `www`. We can remove all of the component imports there once #353 is merged.

PR #364, opened 2019-08-21
fix: don't hide header at any viewport size

vpicone:
@mjabbink This brings the header more in line with the shell guidance:

>The header spans the full width of the viewport and is the topmost element in the browser window. Header elements are persistent throughout the product experience.

[source](https://www.carbondesignsystem.com/components/UI-shell-header/usage#anatomy)

It also pulls the page tabs out of the header so their sticky logic doesn't depend on page header height.


packages/gatsby-theme-carbon/src/components/PageTabs/PageTabs.module.scss
```diff
@@ -1,7 +1,10 @@
 .tabs-container {
   margin: 0;
   width: 100%;
-  box-shadow: 0 -1px 0 rgba($ui-02, 0.25);
```
alisonjoseph: Is this removed on purpose??

packages/gatsby-theme-carbon/src/components/PageHeader/PageHeader.module.scss
```diff
@@ -1,28 +1,19 @@
...
 
 .page-header {
-  background: $carbon--black-100;
+  background: $ui-05;
   color: $inverse-01;
   width: 100%;
   display: flex;
   flex-direction: column;
   justify-content: flex-end;
-  height: 20rem;
```
alisonjoseph: is this height change on purpose?

packages/gatsby-theme-carbon/src/components/PageTabs/PageTabs.module.scss
vpicone: Yeah it was emulating a border top which is now a proper border bottom in the page header.

packages/gatsby-theme-carbon/src/components/PageHeader/PageHeader.module.scss
vpicone: Yeah the tabs are removed

alisonjoseph: It's supposed to be the same height with or without tavs, the 20rem is correct. Unless there is an updated design somewhere??

packages/gatsby-theme-carbon/src/components/PageHeader/PageHeader.module.scss
```diff
@@ -1,28 +1,19 @@
 @import '~carbon-components/scss/globals/scss/layout';
 
 .page-header {
-  background: $carbon--black-100;
+  background: $ui-05;
```
alisonjoseph: Is there a reason we changed this from black to gray-100?

packages/gatsby-theme-carbon/src/components/PageTabs/PageTabs.module.scss
alisonjoseph: 👍

packages/gatsby-theme-carbon/src/components/PageTabs/PageTabs.module.scss
```diff
@@ -1,7 +1,10 @@
...
+  background: $ui-05;
```
alisonjoseph: Same as the above question, should be black not gray-100?

alisonjoseph:
Overall looks good, had a few comments about some design changes in this PR though, not sure where they are coming from??

packages/gatsby-theme-carbon/src/components/PageHeader/PageHeader.module.scss
vpicone: Ah, unintentional. To change the height now, you'll just need to set the height custom property. It'll shorten it by 3rem if there's tabs.

```css
--height: 14rem
```

packages/gatsby-theme-carbon/src/components/PageHeader/PageHeader.module.scss
vpicone: Whoops, was trying to switch to use theme tokens and misread this.

packages/gatsby-theme-carbon/src/components/PageTabs/PageTabs.module.scss
vpicone: same, switched it back

PR #365, opened 2019-08-21
refactor: updates button to div, adds event handlers, and adds video example to gallery

abbeyhrt:
Due to the custom Video button, the videos in the ImageGallery were getting this error: 
Screen Shot 2019-08-21 at 11 34 53 AM

To solve this, this updates the Play/Pause `button` to a `div` and adds the event handlers to handle play/pause on keydown events enter and space. 

Testing: Go to the video in the Image Gallery, verify in the console that the above error isn't occurring. Open the gallery on the image and confirm that you can tab into the Play/Pause button and play/pause on enter or space bar.


vpicone:
Do you think it'd be simpler to make the gallery components divs? Thanks for tackling this!

abbeyhrt:
I think it'd be about the same, really. I get wanting to change the imageGallery instead of the Video though since it's the one creating the conflict and is more of a one-off than the Video.

packages/gatsby-theme-carbon/src/components/Video/Video.js
```diff
@@ -63,22 +63,51 @@ const Video = ({ vimeoId, title, ...props }) => {
       });
   }
 
+  function onVimeoKeyDown(event) {
+    const player = getVimeoPlayer();
+
```
emyarod:
```suggestion
event.preventDefault();
```

abbeyhrt:
@alisonjoseph when you tab through to the video does it skip one and then focus on the video button on the next tab?

packages/gatsby-theme-carbon/src/components/Video/Video.js
abbeyhrt: Added this! Good catch, thank you!

alisonjoseph:
Yea, the focus order is fine, we're just missing the blue outline on the video as you tab through.

abbeyhrt:
@emyarod I confirmed with Alison that the focus is fine as is, the video button receives focus and opens the gallery, would you be able to take a second look at this?

PR #368, opened 2019-08-22
fix: article card type style and caption spacing

alisonjoseph:
- type size broke on article cards w/ latest Carbon - fixed
- updates spacing for captions below video
- token color change broke anchor link color - fixed  ref #369

### DIFF
--- a/packages/gatsby-theme-carbon/src/components/Caption/caption.scss
+++ b/packages/gatsby-theme-carbon/src/components/Caption/caption.scss
@@ -12,7 +12,7 @@ img + .#{$prefix}--caption {
 
 // Caption after video
 div[class^='Video-module--video'] + .#{$prefix}--caption {
-  margin-top: -4rem;
+  margin-top: -$layout-05;
 }
 
 // Responsive by default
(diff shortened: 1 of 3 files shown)


packages/gatsby-theme-carbon/src/components/Caption/caption.scss
```diff
@@ -11,8 +11,8 @@ img + .#{$prefix}--caption {
 }
 
 // Caption after video
-[class^='Video-module--video'] + .#{$prefix}--caption {
-  margin-top: -$spacing-08;
+div[class^='Video-module--video'] + .#{$prefix}--caption {
+  margin-top: -4rem;
```
emyarod: just wondering, we don't have a token for this?

alisonjoseph: Good call, I can change to `$layout-05` we just don't have a spacing token that goes up to 4rem

emyarod: or maybe the mini-units mixin?

PR #373, opened 2019-08-26
Content vertical spacing normalization/refactor

vpicone:
This PR is aimed at exploring a more holistic/consistent approach to styling.

The base vertical rhythm unit (1.5rem or 24px) is based on the base font size 16px * the base line height (1.5). This unit is used as the default margin-top for all block position elements on a page.

Exceptions to the rule come from nested elements, and one-offs (headers, captions).

https://github.com/carbon-design-system/gatsby-theme-carbon/issues/252

To-do:
- [ ] apply styles to nested instances (columns)


alisonjoseph:
This approach seems great, will need to sit down and review w/ Wonil and Jeanne

jeanservaas:
Hey Vince,

I just noticed that the spacing between the intro paragraph and the anchor links looks too small now, it's supposed to be 32px, but right now it's 16px... which doesn't create enough separation. Before I make an issue, Alison suggested I comment here in case it was related to this work.

image

vpicone:
@jeanservaas fixed
@alisonjoseph definitely need to have a fallback for the base spacing. Using the custom property makes for really easy overrides since we can just update `--space` without worrying about specificity:

```css
.main > * + * {
--space: 1.5rem;
}

/* Technically less-specific selector,
but since it's closest to the element it's 
value used instead of the default */
h4 {
--space: 0; 
}
```

I think losing the overrides in IE is worth the benefit of how easy it is to override without specificity wrestling.

rcwakelin:
@alisonjoseph @vpicone checking in....

rcwakelin:
@mattrosno adding you to the chain

jeanservaas:
I talked to @mjabbink yesterday and both of us do think that this is better and that we should merge this in just to level-set our vertical spacing and clean the shizz up. 

Our biggest issue, is going to be the H4... both margin top and margin bottom

image

**H4 Margin bottom (intended):**

image


**H4 Margin bottom (not intended):**

image

**H4 Margin top (good in the flow of paragraphs):**

image

**H4 Margin top (not good sandwiched between components or images):**

image

PR #379, opened 2019-08-27
Fix/active in nav

SilverFox70:
We are still experiencing issues with duplicate selection of similar titled items (Issue #265) in the nav. This fix works for us locally.


vpicone:
@SilverFox70 it looks like this is fixed on ibm.com/brand, is this the implementation you're using there?

SilverFox70:
Thanks for reviewing @vpicone , I'll loop back with an explanation tomorrow!

SilverFox70:
@vpicone when using the following code
```
return (
    <Location>
      {({ location }) => {
        const pathname = pathPrefix
          ? location.pathname.replace(pathPrefix, '')
          : location.pathname;

        const isActive = items.some(
          item => pathname === item.path || pathname === `${item.path}/`
        );
...
```
We see an issue with tabbed pages. Notice below how we are on a page and yet the nav is not showing anything active:
Screen Shot 2019-09-19 at 9 59 23 AM

But if instead I use the following:
```
<Location>
      {({ location }) => {
        const pathname = pathPrefix
          ? location.pathname.replace(pathPrefix, '')
          : location.pathname;

        const isActive = items.some(item =>
          pathname.split('/').length > 2
            ? item.path.slice(0, item.path.lastIndexOf('/')) ===
              pathname.slice(0, pathname.lastIndexOf('/'))
            : item.path.replace(/\/$/, '') === pathname
        );
...
```
Then the Left Nav shows the active state correctly:
Screen Shot 2019-09-19 at 9 59 59 AM

packages/gatsby-theme-carbon/src/components/LeftNav/LeftNavItem.js
```diff
@@ -24,17 +24,14 @@ const LeftNavItem = props => {
         const pathname = pathPrefix
           ? location.pathname.replace(pathPrefix, '')
           : location.pathname;
-        const isActive = items.some(
-          item => pathname === item.path || pathname === `${item.path}/`
+
```
vpicone: Ah! That totally makes sense, thanks again for your patience with this, just wanted to add a comment for if we have to revisit this logic down the line.

```suggestion
// If item a dropdown, ignore the last path segment to account for tabbed
// items, which won't show up in the items array. For non-dropdowns,
// just check for exact path match while removing trailing slash.
```

SilverFox70:
@vpicone I accepted your suggestion for the comment. Any idea on when this might get merged in and then released?

vpicone:
@SilverFox70 oh wow I have no idea how this one slipped through the cracks It looks good to me!

PR #399, opened 2019-09-04
fix(Code): align code blocks in lists; closes #398

### LINKED ISSUE #398: fenced code block within list misaligned
Here's a screenshot

image

The code block is within the `<ul>`, but it should be aligned to the text ("Arguments", "Returns", "Example").

boneskull:
This "works" but may not be "correct", as I'm unclear about the intention of the margins with this thing.  Responsiveness seems OK.

closes #398


alisonjoseph:
Going to defer to design on this question, as I believe the intent is for the code blocks to bleed into the gutter on the left edge. However, not sure about when in a list. @jeanservaas 

https://gatsby-theme-carbon-git-fork-boneskull-code-in-list.carbon-design-system.now.sh/components/code-blocks

Screen Shot 2019-09-05 at 7 06 02 AM

boneskull:
I’m not a designer, but I’d expect the left-alignment to be consistent within a list (like gfm rendering).

alisonjoseph:
@jeanservaas any feedback on this? (how code blocks should align when within a list)

boneskull:
@alisonjoseph Sorry, I don't understand; are you asking me for a fix?  I will need some hand-holding if I'm to change the right margin.  I originally tried to change this myself, and couldn't figure it out.

alisonjoseph:
I'll try and take a look tomorrow and see what we need to do to fix the margins here. 👍

packages/gatsby-theme-carbon/src/components/markdown/Markdown.module.scss
```diff
@@ -64,8 +64,7 @@
 
 // Fix bleed when code snippets are used in a list
 .list-item > pre {
-  margin-left: -1rem;
-  margin-right: -1.5rem;
+  margin-left: 1rem;
   margin-top: $spacing-05;
   @include carbon--breakpoint('md') {
     margin-right: 1.5rem;
```
alisonjoseph:
```suggestion
    margin-right: -$spacing-05;
```

This should fix the right alignment!

boneskull:
@alisonjoseph hmm, not quite...

image

alisonjoseph:
I think that’s a bug with the code snippet being slightly off grid. I can open a separate PR to fix that.

boneskull:
I’ll take your word for it! LGTM then, fwiw.

boneskull:
@alisonjoseph thanks!

PR #400, opened 2019-09-04
feat(ResourceCard): add a blue background variation

sstrubberg:
Updated the ResourceCard component to include a blue60 background with a blue70 hover.

### DIFF
--- a/packages/example/src/pages/components/ResourceCard.mdx
+++ b/packages/example/src/pages/components/ResourceCard.mdx
@@ -41,7 +41,7 @@ The `<ResourceCard>` component should be wrapped with a `<Column>` inside of `<R
       subTitle="Alternate color"
       title="Dark"
       aspectRatio="2:1"
-      color="dark"
+      color="blue"
       actionIcon="email"
       href="https://github.com/IBM/carbon-elements/blob/master/.github/CONTRIBUTING.md"
       >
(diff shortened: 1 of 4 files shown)


sstrubberg:
Made some styles for disabled and soften the top blue border.

image

We are currently using it as a way to draw out a "main" call to action.

image

We experimented with using the `dark` option, but that gave us an almost "disabled" feeling when we used it.

jeanservaas:
@sstrubberg 

Apologize for the delay, I just saw this. The blue looks very aggressive, especially in the light theme and it's a color that's usually reserved for our buttons, not our resource cards, which are more akin to links. 

I'm wondering 1) why this card needs emphasis and 2)  if you can try some other exploration if you need emphasis here. Maybe scale rather than color. We've actually been exploring some smaller resource tiles for certain situations we've encountered on Carbon. But I don't think we're ready to move forward with this as a variation yet.

sstrubberg:
> @sstrubberg
> 
> Apologize for the delay, I just saw this. The blue looks very aggressive, especially in the light theme and it's a color that's usually reserved for our buttons, not our resource cards, which are more akin to links.
> 
> I'm wondering 1) why this card needs emphasis and 2) if you can try some other exploration if you need emphasis here. Maybe scale rather than color. We've actually been exploring some smaller resource tiles for certain situations we've encountered on Carbon. But I don't think we're ready to move forward with this as a variation yet.

Thanks for the feedback @jeanservaas. We want to draw emphasis to the card because we have a block of four cards with various links to resources. We want to elevate one of the four cards as our main call-to-action and we felt like the blue60 was doing a pretty solid job compared to the dark mode tile that is currently in the theme. We'll do some exploring into how we might elevate the CTA without relying on the blue. We'll also do some testing with users to see if the blue comes off as too aggressive.

mjabbink:
No blue resource cards please

PR #404, opened 2019-09-06
feat: add prepend option to titleType

### LINKED ISSUE #403: Need "prepend" titleType
Need titles to be `page name - site name` `Button - Carbon Design System` ref https://github.com/carbon-design-system/carbon-website/issues/262

alisonjoseph:
closes #403 

Adds option to prepend the title before the site name 

`prepend`: "Configuration - Gatsby Theme Carbon"

### DIFF
--- a/packages/example/src/pages/guides/configuration.mdx
+++ b/packages/example/src/pages/guides/configuration.mdx
@@ -170,6 +170,7 @@ To add a link to the bottom of each page that points to the current page source 
   - `page`: "Configuration" (default)
   - `site`: "Gatsby Theme Carbon"
   - `append`: "Gatsby Theme Carbon – Configuration"
+  - `prepend`: "Configuration - Gatsby Theme Carbon"
 
 ```js
 plugins: [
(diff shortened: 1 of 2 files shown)


packages/example/src/pages/guides/configuration.mdx
```diff
@@ -170,6 +170,7 @@ To add a link to the bottom of each page that points to the current page source
   - `page`: "Configuration" (default)
   - `site`: "Gatsby Theme Carbon"
   - `append`: "Gatsby Theme Carbon – Configuration"
+  - `prepend`: "Configuration - Gatsby Theme Carbon"
```
jnm2377: Looks good. Should we update line 168 to be "Here are the _four_ options..."?

> - `titleType` – pick between three formats for the `<title>` element for your site. Here are the three options using this page as an example:

alisonjoseph: oh yea, good catch

packages/gatsby-theme-carbon/src/components/Meta.js
```diff
@@ -14,6 +14,8 @@ const Meta = ({ pageTitle, pageDescription, pageKeywords, titleType }) => {
         return title;
       case 'append':
         return `${title}${pageTitle ? ` – ${pageTitle}` : ''}`;
+      case 'prepend':
+        return `${pageTitle}${pageTitle ? ` – ${title}` : ''}`;
```
vpicone: We probably want to flip this so in the event there’s no page title we still have a title. This is a factor for things like the homepage and 404. 

The hyphen should be attached to the pageTitle so we don’t have a dangling hyphen in that scenario as well.

PR #410, opened 2019-09-11
[Search] add tab to distinguish pages with the same title

vpicone:
Pages with tabs could share the same title. If this is the case, we can add the last bit of path as an identifier to set them apart.

`→` character got okayed by @designertyler 

**You can search for “Navigation” to see it working**

### DIFF
--- a/packages/gatsby-theme-carbon/src/components/GlobalSearch/Menu.js
+++ b/packages/gatsby-theme-carbon/src/components/GlobalSearch/Menu.js
@@ -7,6 +7,7 @@ import {
   description,
   active,
   hidden,
+  tab,
 } from './GlobalSearch.module.scss';
 
 export const MenuContext = createContext();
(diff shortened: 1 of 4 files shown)


packages/gatsby-theme-carbon/src/components/GlobalSearch/Menu.js
```diff
@@ -69,10 +87,11 @@ const MenuItem = ({ page, index, onKeyDown, id }) => {
         className={className}
         to={page.path}
       >
-        {`${page.title} – `}
-        <span className={description}>
-          {page.description && page.description.toLowerCase()}
-        </span>
+        {page.title}&nbsp;
```
abbeyhrt: What is the `&nbsp` doing? Nothing wrong with it! I just have no idea what that is

vpicone: It stands for non-breaking-space, it's just an alternative to the {" "} weirdness of JSX for introducing spaces.

abbeyhrt: ah okay, thank you!

PR #420, opened 2019-09-12
feat: export custom Title component from MDX

### LINKED ISSUE #409: Support for choosing line break position in page-title
Aesthetically this has some intersection with #308 and also relates to Brand Center #443. The desire is to be able to finesse where a line break occurs in the page-title as it appears in the PageHeader. We did a brief exploration of achieving this:
```
<h1 id="page-title" className={text}>
  <div dangerouslySetInnerHTML={{ __html: frontmatter.title || null }}/>
</h1>
```

vpicone:
closes #409 

I think this stands to be a particularly powerful pattern for things like theming individual pages if we identify future components where we _want_ to permit that sort of deviation.

### DIFF
--- a/packages/example/src/pages/components/MDX.mdx
+++ b/packages/example/src/pages/components/MDX.mdx
@@ -25,7 +25,7 @@ layout.
 ## Frontmatter
 
 You can declare frontmatter in your `.mdx` files to provide specific metadata for the theme to use.
-The most important of which is the `title`. You can also provide a descriptions and keywords which will be added to the `head` of your document.
+The most important of which is the `title`. You can also provide a description and keywords which will be added to the `head` of your document.
 
 ```md
 ---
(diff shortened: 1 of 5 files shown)


packages/example/src/pages/components/MDX.mdx
```diff
@@ -0,0 +1,55 @@
...
+
+<AnchorLinks>
+  <AnchorLink>Frontmatter</AnchorLink>
+  <AnchorLink>Custom title</AnchorLink>
+</AnchorLinks>
+
+## Frontmatter
+
+You can declare frontmatter in your `.mdx` files to provide specific metadata for the theme to use.
+The most important of which is the `title`. You can also provide a descriptions and keywords which will be added to the `head` of your document.
```
emyarod:
```suggestion
The most important of which is the `title`. You can also provide a description and keywords which will be added to the `head` of your document.
```

PR #429, opened 2019-09-17
feat: add InlineNotification component

### LINKED ISSUE #427: Feat: Low-contrast notification for adding notes to guidance
### Design & spec

image

### In situ: 

**Note** it hangs left into gutter.

image

alisonjoseph:
Closes #427 

Adds inline notification component, uses `children` instead of the title/subtitle props directly from the Carbon component to allow for markdown (links) inside of the notification.

### DIFF
--- a/packages/gatsby-theme-carbon/src/components/InlineNotification/InlineNotification.js
+++ b/packages/gatsby-theme-carbon/src/components/InlineNotification/InlineNotification.js
@@ -31,7 +31,6 @@ export default class InlineNotification extends React.Component {
             lowContrast
             hideCloseButton
             kind={kind}
-            title=""
             subtitle={children}
           />
         </Column>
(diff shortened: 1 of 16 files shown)


shixiedesign:
Nice! Though, it looks like the drop shadow is still here and outline is missing. Did your custom styling not load? Checked in Chrome both regular and incognito. 

image

alisonjoseph:
@shixiedesign weird! the local build is working but not the preview build. I'll take a look

alisonjoseph:
should be fixed now!

alisonjoseph:
Should be ok now

packages/gatsby-theme-carbon/src/components/InlineNotification/InlineNotification.js
```diff
@@ -0,0 +1,40 @@
...
+    className: PropTypes.string,
+    kind: PropTypes.oneOf(['error', 'info', 'success', 'warning']),
+  };
+
+  static defaultProps = {
+    kind: 'info',
+  };
+
+  render() {
+    const { children, className, kind } = this.props;
```
vpicone: instead of pulling kind out can we spread the props into the component as the last prop? That way we'd handle all the other props/future props on the Carbon component.

alisonjoseph: I can, but not sure we want all those options available here, since its a little different use-case than the Carbon component? We always want the close button hidden for example, and always the lower contrast.

PR #444, opened 2019-09-24
feat: rename baseUrl to url; closes #443

### LINKED ISSUE #443: enable better integration with repository field of a consumer's package.json
# Summary

Rename the theme option `repository.baseUrl` to `repository.url`.

### Justification

It'd enable the following.  Given a `package.json` containing:

```json
{
  "repository": {
    "url": "https://github.com/some/where",
    "type": "git",
    "subDirectory": "if/needed"
  }
}
```

A site consuming this theme could, in `gatsby-config.js`,

```js
const {repository} = require('./package.json');

module.exports = {
  plugins: [
    {
      options: {repository}
    }
};
```

This obviously has little material impact, since the workaround is trivial, but would be "nice" regardless.

Ensure that use of the `baseUrl` option doesn't break; prefer `url` but also support a deprecated `baseUrl`.  Could eventually be dropped entirely.

### Available extra resources

I'm happy to send a PR to implement.

boneskull:
Closes #443

Renames the `repository.baseUrl` theme option to `repository.url` for ease of integration with `package.json`.

#### Changelog

**Changed**

- `baseUrl` is now considered deprecated, but backwards compatibility is retained for existing configurations and `EditLink` components
- `url` will no longer appear in `siteMetadata` of graphql schema
- no deprecation warning is emitted
- fixed a bug when using `EditLink` component without a `subDirectory` property, which would cause the string `undefined` to appear in the "Edit on GitHub" link's `href` attribute
- updated documentation
- updated example


boneskull:
You can test the efficacy of the backwards compat by:

1. Providing `repository.baseUrl` (instead of `url`) in the example's `gatsby-config.js`
1. Adding `repository={{baseUrl: 'http://whatever'}}` to the `EditLink` component on [L52 of `src/templates/Default.js`](https://github.com/carbon-design-system/gatsby-theme-carbon/blob/fc1fc02e9a855992dc20d9248537730eda50e48b/packages/gatsby-theme-carbon/src/templates/Default.js#L52)

You can assert that this change fixes a bug by adding `repository={{baseUrl: 'http://whatever'}}` to the same `Default.js` in `master` branch, where the rendered link will point to `http://whatever/undefinedsrc/pages/whateverpage.md`

boneskull:
What's unclear to me about Gatsby is if the graphql "source data" can be persisted in a way that the `siteMetadata` could still contain `baseUrl`, even after an upgrade--maybe living in `.cache` somewhere.  But I'm pretty sure an upgrade of the theme would cause `gatsby` to blast the cache and start from scratch.

boneskull:
resolved conflicts

vpicone:
I disagree, I think `baseUrl` suggests that it's not the entirety of the URL (which is true, considering user's can supply a sub-directory). If the parameter were changed to `url`, it could be misconstrued as the full url to the site including subdir.

I'd rather not deprecate something/create an awkward, future breaking for something as small as this. These things can add up over time. Gonna close this for now.

PR #447, opened 2019-09-25
feat: allow granular control over lines to copy

boneskull:
Sometimes you might not want to copy the entire code block.  For instance, given a shell command:

````
```bash
$ some-command
a bunch of useless output
lkasjdflskjdf
duh duh duh
```
````

When the 'copy' button is clicked, it makes more sense to copy _only_ the command, instead of the command's output.

This PR adds `copyLineStart` and `copyLineLength` props to code blocks for more granular control over what exactly gets copied when the button is clicked.

#### Changelog

**New**

- Adds `copyLineStart` and `copyLineLength` props to code blocks for control over which lines get copied.

**Changed**

- Updated example (and some reformatting, apparently)

### CODE REVIEW

**Comment by boneskull** [comment]
@vpicone 

> Do we need all the parse int guarding/take both strings and numbers since we know the source is coming from mdx meta?

AFAICT we're always going to get strings out of MDX, so we need `parseInt` in some fashion.  I could be mistaken, but the value of `foo=true` is a `string`...

> Would prefer we return early rather than the wonky ternary at the end.

I can do that, but please keep in mind: less experienced contributors might take offense to "wonky" :wink:

> Can we use parameter destructuring instead of the || short circuits.

These are intentional.  Destructuring will provide a default for an `undefined` argument, but not a falsy one; a falsy argument should also be converted to the default (essentially preferring the default over `NaN` and bailing).  To make it more clear, I could add comments or otherwise .. uh.. increase verbosity.

**Comment by boneskull** [comment]
I've pushed an update which should make the function a little more clear

**Comment by boneskull** [comment]
So, this is relevant.  I read this article just yesterday regarding code samples:

https://stoplight.io/blog/writing-documentation-when-you-arent-a-technical-writer-part-one-ef08a09870d1/#code-samples

The gist of this is, "make code easy to copy".  The article calls out the convention of displaying terminal commands with a leading `$`, like so:

```bash
$ ls *.js
foo.js       bar.js
```

With this change, I can control which line(s) are copied, which is good.  What I _can't_ do is omit the leading `$ `, so a reader could more easily copy/paste the command itself.  

I'd like to do something about this, and there are several ways to go about it.  Curious what you (@vpicone) and @alisonjoseph think.  Here are a few ideas:

1. Detect `shell` or `bash` language. When copying, omit leading `$ ` from each line copied.  This respects the convention, but in rare cases may not be what a consumer wants to happen.
1. Allow granular control over not just which _lines_ to copy, but at which character position the first line (and ending line, for consistency) begins copying.
1. Detect leading `$` and use CSS trickery to omit it from the code block, yet display it like it's within the block.  I wouldn't know the best way of going about this.
1. Maybe [Prism](https://prismjs.com/) has some facility to do this?  It's not a new problem.

boneskull:
I think what I'm missing is an understanding of project scope: what's out-of-scope vs. what do we feel is important to provide out-of-the-box for consumers?

If this theme is adopted by more teams building technical project documentation, then this theme should provide a full-featured Code component.  I've added other features to my own shadowed copy, but am holding off on sending any PRs until I understand this. 
 
I'm happy to propose more ideas, if you're interested & available to consider them.

What are the use-cases of the current base of consumers? Am I the only one building technical docs with it (this, specifically: https://ibm.github.io/report-toolkit) instead of a marketing site?  Is "technical documentation" something this theme wants to better support?

vpicone:
@boneskull The carbon design system site is using the theme for documentation. I'm not sure that only copying a subset of lines is expected ux for a copy button so gonna close this for now. Definitely encourage you to shadow this behavior if you feel like you'd like to provide it for your site.

PR #454, opened 2019-09-26
fix: accordion padding

sstrubberg:
Added 4rem of padding-right to accordion per [Mike's request](https://github.com/IBM-Design/idr/issues/133)

image


mjabbink:
looks good. @alisonjoseph and Jeannie Servaas (can’t ping in here)

mjabbink:
looks right

PR #468, opened 2019-10-09
fix(ImageCard): add focus border

### LINKED ISSUE #432: ImageCard is missing focus state
Needs 2px outline focus state that is present on the other card components.

jnm2377:
Closes #432 

Adds focus border to `ImageCard`

#### Changelog

**New**

- added `::after` styles to tile. 
- adds devDep `"eslint-config-react-app": "^5.0.2",` => this was because this is being hoisted in the Gatsby dependency so it's not at the top level in node_modules and my local env was breaking bc of it. Tried a couple of different ways to fix it, but with no real luck, so @joshblack and I decided this is probably the easiest fix (also in case other people run into this issue in the future).

**Reviewers**
_Note: I tried doing `outline`, `outline-offset`,  `border` and `box-shadow` as alternatives to focus the focus state but none of them were adding the styles properly. Adding the `::after` was a last resort. It seems that we do that already for `ImageCard`, so I thought it would be ok. Tile seems to be working the same when it is focused + hovered, focused + clicked._
- ensure tile is still working as expected
- ensure focus style looks correct

### DIFF
--- a/packages/gatsby-theme-carbon/src/components/ImageCard/image-card.scss
+++ b/packages/gatsby-theme-carbon/src/components/ImageCard/image-card.scss
@@ -3,11 +3,34 @@
   padding: $spacing-05 25% $spacing-05 $spacing-05;
   position: relative;
   text-decoration: none;
+
+  &::after {
+    position: absolute;
+    top: 2px;
+    left: 2px;
+    content: '';
+    width: calc(100% - 4px);
+    height: calc(100% - 4px);
+    background: none;
+    outline: 2px solid transparent;
+    transition: outline motion(standard, productive) $duration--fast-01;
+  }
 }
 
 .#{$prefix}--image-card .#{$prefix}--tile:focus {
   outline: none;
-  //need to fix/add focus state
+
+  &::after {
+    position: absolute;
+    top: 2px;
+    left: 2px;
+    content: '';
+    width: calc(100% - 4px);
+    height: calc(100% - 4px);
+    background: none;
+    outline: 2px solid $focus;
+    transition: outline motion(standard, productive) $duration--fast-01;
+  }
 }
 
 .#{$prefix}--image-card__title {
--- a/packages/gatsby-theme-carbon/src/components/ImageCard/image-card.scss
+++ b/packages/gatsby-theme-carbon/src/components/ImageCard/image-card.scss
@@ -6,11 +6,11 @@
 
   &::after {
     position: absolute;
-    top: 2px;
-    left: 2px;
+    top: rem(2px);
+    left: rem(2px);
     content: '';
-    width: calc(100% - 4px);
-    height: calc(100% - 4px);
+    width: calc(100% - rem(4px));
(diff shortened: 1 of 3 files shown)


jeanservaas:
I think we should go 3px for focus boarder instead of 2px here to beef it up. It's getting lost.

packages/gatsby-theme-carbon/src/components/ImageCard/image-card.scss
```diff
@@ -3,11 +3,34 @@
...
+  &::after {
+    position: absolute;
+    top: rem(3px);
+    left: rem(3px);
+    content: '';
+    width: calc(100% - 7px);
+    height: calc(100% - 7px);
+    background: none;
+    outline: 3px solid transparent;
+    transition: outline motion(standard, productive) $duration--fast-01;
```
emyarod: rather than the 3px `top` and `left` we can just use `outline-offset`, and rather than the `100% - 7px` for `width` and `height` calculation we can just reduce the dimensions by 1px (because of https://github.com/carbon-design-system/gatsby-theme-carbon/blob/master/packages/gatsby-theme-carbon/src/components/ImageCard/image-card.scss#L62-L63). I know the width and height offset value is 7 to account for 2 sides * 3px + 1px from the gap but that might not be obvious at first glance, so I figured keeping it at 1px in reference to L62-63 might be clearer

```scss
&::after {
  position: absolute;
  top: 0;
  left: 0;
  content: '';
  width: calc(100% - 1px); // 1px to offset negative `left` positioning
  height: calc(100% - 1px); // 1px to offset negative `top` positioning
  outline: 3px solid transparent;
  outline-offset: -3px;
  transition: outline motion(standard, productive) $duration--fast-01;
}
```

packages/gatsby-theme-carbon/src/components/ImageCard/image-card.scss
```diff
@@ -3,11 +3,34 @@
...
+    position: absolute;
+    top: rem(3px);
+    left: rem(3px);
+    content: '';
+    width: calc(100% - 7px);
+    height: calc(100% - 7px);
+    background: none;
+    outline: 3px solid $focus;
+    transition: outline motion(standard, productive) $duration--fast-01;
+  }
```
emyarod: we can dedupe the rules from the non-focused state

```scss
&::after {
  outline: 3px solid $focus;
}
```

vpicone:
@jnm2377 did we still need the eslint config here or was that from some testing?

jnm2377:
@vpicone the eslint config was for the weird bug that was happening for the dependency. Not sure how else to solve it, but it seems that it's not just me who is running into the issue, so I thought it would be easier to include it in case other people have the same issue. Can remove if we don't want to.

alisonjoseph:
@vpicone I actually just added that esling config to master, it was blowing up everywhere for everyone.

jnm2377:
@emyarod updated 👍

PR #469, opened 2019-10-09
feat(EditLink): Added a timestamp

sstrubberg:
I added a timestamp to the EditLink.

image

I'd love some feedback on this. I'd like to eventually have the timestamp be an option where it's disabled by default even if you choose to enable the EditLink.


vpicone:
We'd likely need some design input on this

jeanservaas:
Hey @sstrubberg thanks for this, just a couple of suggestions.

Right now the type looks a little disjointed because of the line breaks. I suggest keeping the lines together $body-long-01 (with 20px line height) and making the "Edit this link" the Gray 100 and making the date stamp the Gray 70 so the link takes priority in hierarchy.

image

mjabbink:
Why are we pulling "Last updated..." out of the footer?

mjabbink:
Or is it in both places? Is that necessary?

sstrubberg:
> Or is it in both places? Is that necessary?

I didn't pull it out of the footer. I thought having the time stamp would serve two purposes: 1) Let's people know up-front when things were last updated instead of being buried in the footer and 2) By having the timestamp next to the "Edit on Github" link, I think it further motivates a person to make changes to the page if they see something broken or out of date.

mjabbink:
Makes sense. Then Jeannie’s recommendation is a good one

sstrubberg:
image

Made the edit's per Jeannie's recommendations.

packages/gatsby-theme-carbon/src/components/EditLink/EditLink.module.scss
```diff
@@ -1,7 +1,7 @@
 .link {
   @include carbon--type-style('body-short-01');
   transition: color $transition--base;
-  color: $text-02;
+  color: $carbon--gray-100;
```
alisonjoseph:
```suggestion
  color: $text-01;
```

chrisconnors-ibm:
> Why are we pulling "Last updated..." out of the footer?

@sstrubberg are they distinct values? ie footer = last website change/deploy, while this is last time the _page itself_ was edited?

sstrubberg:
> > Why are we pulling "Last updated..." out of the footer?
> 
> @sstrubberg are they distinct values? ie footer = last website change/deploy, while this is last time the _page itself_ was edited?

Unfortunately, it's not "page-specific". From what I can tell, there is no timestamp standard in the footer. If you're seeing it on a particular site, it might be using the same library I'm using or it's being updated manually. (the IDR site is manually edited currently)

chrisconnors-ibm:
Sorry @sstrubberg - on our site I'm pretty sure in our case the footer 'Last updated date" = the last deploy. We've wanted to _also_ show the last edit date for the pages with an "Edit this page on GitHub" link to show when that Markdown was last committed (which is likely very different from last _deploy_. Is that your intent here? either way that's something we've been looking for. Last edit date gives us a measurement of freshness/staleness of a specific page.

@mjabbink - that's why they'd be different values.

sstrubberg:
> Sorry @sstrubberg - on our site I'm pretty sure in our case the footer 'Last updated date" = the date of most recent deploy. We've wanted to _also_ show the last edit date for the pages with an "Edit this page on GitHub" link to show when that Markdown was last committed (which is likely very different from last _deploy_. Is that your intent here? either way that's something we've been looking for. Last edit date gives us a measurement of freshness/staleness of a specific page.
> 
> @mjabbink - that's why they'd be different values.

Totally hear you. I did some investigating, turns our the Carbon site does indeed [have a timestamp](https://github.com/carbon-design-system/carbon-website/blob/acad8caae9654b468daa011a67e6c139ded17785/src/gatsby-theme-carbon/components/Footer.js#L15) that is created in the footer. This, however, is unique to the Carbon site and is not in the Gatsby Theme (it probably should though!). The component I whipped up for the page uses a very similar tactic to what Carbon's is using, by timestamping based on the site's build time. I haven't figured out how to check the modified date on an individual MDX file yet.

vpicone:
@sstrubberg sorry I let this one slip through the cracks! A few issues:

1) I agree with @chrisconnors-ibm: having the global site-build time next to the "edit in github" link is misleading. It gives the impression that _that particular page_ was edited recently.

2) If you don't specify a specific time, the Moment library will use the browser time. This means when you go to the page, it'll show the current date (not what you intended I imagine haha)

3) Second, we can totally do this without a dependancy there's [Intl.DateTimeFormat](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/DateTimeFormat) built into javascript.

Fortunately we can solve all three of these problems in one go! Try spinning spinning up the theme dev enviroment and head to http://localhost:8000/___graphql

Then paste in this query:

```grapqhl
{
  allMdx {
    edges {
      node {
        parent {
          ... on File {
            mtime(formatString: "MMMM Do, YYYY")
          }
        }
        frontmatter {
          title
        }
      }
    }
  }
}
```

We can get the `mtime` (modified time) of the actual MDX file. This would normally be a PageQuery in Gatsby but those are difficult with MDX. Try digging around in `onCreatePage` of `gatsby-node.js` in the theme.

sstrubberg:
@vpicone took me to school this morning and we got it working where every MDX page that is generated will get a timestamp based on the last time it was edited by a local machine.

https://gatsby-theme-carbon-git-fork-sstrubberg-timestamp.carbon-design-system.now.sh/

sstrubberg:
Hey @vpicone, it looks like it's breaking when it gets deployed. I'm curious if the mtime is getting blown away when zeit builds it?

vpicone:
@sstrubberg definitely a posibility. We're gonna have to find some other way to track it. Maybe we could have a piece of front matter that gets changed when someone makes a PR?

vpicone:
Gonna close this for now, can dig into it again in the new year with a dif approach

PR #480, opened 2019-10-25
chore(Docs): Dark Mode

sstrubberg:
Added documentation to turn on dark mode.

image

Also fixed a link on the configuration page.

### DIFF
--- a/packages/example/src/pages/components/MDX.mdx
+++ b/packages/example/src/pages/components/MDX.mdx
@@ -31,9 +31,16 @@ The most important of which is the `title`. You can also provide a description a
 ---
 title: Markdown
 description: Usage instructions for the Markdown component
+theme: dark
 keywords: 'ibm,carbon,gatsby,mdx,markdown'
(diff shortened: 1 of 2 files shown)


alisonjoseph:
I don't think design wanted us to publish this yet until its ready for primetime.

sstrubberg:
> @sstrubberg I'm OK with this being added with the notification, but I just wanted to ask about the code snippet theming...are you using the Gatsby component? Just wondering why the red highlight. Also, I added @connor-leech just because this is technically content too.

Yep, using the Gatsby Code snippet. I thought that was weird too, but it's what ya'll have in there currently, so i kept it.

vpicone:
@sstrubberg we kinda did a 'soft release' of this, intentionally leaving it out of the docs. It's likely going to break in the future (different strategy, selectors and theme options) and we don't wanna get folks stuck.

We'll definitely add this once we can effectively theme the four main Carbon themes.

PR #485, opened 2019-10-29
fix: text-decoration on links

vpicone:
We use Carbon’s link component but we can add underline ourselves without waiting for approval [upstream in the Carbon library](https://github.com/carbon-design-system/carbon/issues/3908)

Adding it upstream might come as a surprise to folks who use links in menu’s and now they suddenly have underlines.

This is a stop gap measure to add text decoration in the mean time.


vpicone:
@jeanservaas do you mean for Carbon or the theme? I think we should just stop using the carbon link if we’re gonna change it like that probably. We have to think about focus and visited states too

jeanservaas:
talked to @vince face to face! All states remain the same except we're adding a hover... hover used to be the underline 

These are the old Carbon link specs... now the only difference is that hover will be $hover-primary-text and we need to change the value of that token from Blue 70 to Blue 80 in the .json file.

image

vince:
> talked to @vince face to face! All states remain the same except we're adding a hover... hover used to be the underline

Wrong Vince - please stop summoning me.

mjabbink:
Happy to see. I assume IDL and other sites will get this when they update?

mjabbink:
Nice! Works and looks better to me as well.

PR #506, opened 2019-11-06
fix(accordion): add right padding to title

jnm2377:
Rel: https://github.com/carbon-design-system/gatsby-theme-carbon/pull/454

**New**

- adds right padding to the accordion title to match the content per Mike's request

### DIFF
--- a/packages/gatsby-theme-carbon/src/components/Accordion/Accordion.module.scss
+++ b/packages/gatsby-theme-carbon/src/components/Accordion/Accordion.module.scss
@@ -9,6 +9,7 @@
 
 .accordion :global(.bx--accordion__title) {
   @include carbon--type-style('heading-02');
+  padding-right: 4rem;
 }
 
 .accordion :global(.bx--accordion__content) p {
--- a/packages/gatsby-theme-carbon/src/components/Accordion/Accordion.module.scss
+++ b/packages/gatsby-theme-carbon/src/components/Accordion/Accordion.module.scss
@@ -9,7 +9,7 @@
 
 .accordion :global(.bx--accordion__title) {
   @include carbon--type-style('heading-02');
-  padding-right: 2rem;
+  padding-right: $spacing-07;
 }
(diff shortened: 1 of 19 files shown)


mjabbink:
If I click on the preview link above the width still look the same and expand 100%.

Screen Shot 2019-11-13 at 8 39 38 AM

jeanservaas:
@Alison might be easier if I swing by your desk sometime later this week and look at this... I'm not quite sure what the issue is.

jeanservaas:
ok it took me a second, but I get it... @Alison, I think what you have shown above is fine -- if we give the paragraph the same right margin as the title for the Gatsby theme component... it's going to look weird, because the accordion paragraph will be shorter than the other body copy.  

right now the paragraphs within the accordion and outside of the accordion have the same line length and I think that's good. I also think it's ok that on the off chance you get a really long two-line accordion title, the right padding is slightly larger to avoid the icon.

image

jnm2377:
So should I leave the padding as is or change it? @jeanservaas

jeanservaas:
@jnm2377 what was mike's ask? what was the original issue... not clear what he wants

jnm2377:
Updated title padding to be aligned with content @jeanservaas @alisonjoseph 

Screen Shot 2019-11-18 at 1 36 19 PM
Screen Shot 2019-11-18 at 1 36 38 PM

packages/gatsby-theme-carbon/src/components/Accordion/Accordion.module.scss
```diff
@@ -9,6 +9,7 @@
 
 .accordion :global(.bx--accordion__title) {
   @include carbon--type-style('heading-02');
+  padding-right: 2rem;
```
alisonjoseph:
```suggestion
  padding-right: $spacing-07;
```

mjabbink:
Looks right to me now too

PR #515, opened 2019-11-11
fix(image-card): render as div if no href is provided

### LINKED ISSUE #514: [ImageCard] when no href prop provided, the ImageCard renders an anchor without a href
## Detailed description

Currently if you don't provide an `href` prop to the `ImageCard` component, it will render as an anchor without an `href`. Note that the `href` prop is not required.

If no `href` is provided, I would expect the `ImageCard` to render as a `div` instead (similar to when you mark an `ImageCard` as `disabled={true}`).

jendowns:
Closes #514 

Currently unless you add the `disabled` prop, an `ImageCard` will always render as an anchor.
This is problematic if you don't also provide an `href` (which is not a required prop, either).

Recently on the IBM Security design team, we've come across a situation where we want the positioning and styling of an `ImageCard` within a grid, but we don't need to link it anywhere. (So, we aren't providing an `href`). The `ImageGalleryCard` would not apply well for our designs/situation., and we also don't want to apply `disabled={true}` to `ImageCard` which would apply "do not click" cursor + icon + give an impression that the Card would be clickable under other circumstances. So we wanted to propose a fix here.

Please let me know what you think -- thanks!

#### Changelog

**Changed**

- only apply `bx--tile--clickable` class if `href === undefined` (for hover effects on links only, etc)
- render `ImageCard` as a `div` if `href === undefined`

### DIFF
--- a/packages/gatsby-theme-carbon/src/components/ImageCard/ImageCard.js
+++ b/packages/gatsby-theme-carbon/src/components/ImageCard/ImageCard.js
@@ -101,6 +101,8 @@ export default class ImageCard extends React.Component {
     let isLink;
     if (href !== undefined) {
       isLink = href.charAt(0) === '/';
+    } else {
+      isLink = false;
     }
 
     const ImageCardClassNames = classnames([`${prefix}--image-card`], {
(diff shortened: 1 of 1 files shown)


SimonFinney:
For context, we're currently migrating our experience guide over to use `gatsby-theme-carbon` under the hood - we'd like to reuse as much as possible and have several instances in which we use `ImageCard` with no link out to somewhere else.

As a workaround, we thought we could linking directly to the static image but this doesn't play good with Gatsby.

PR #518, opened 2019-11-13
fix(feature-card): don't provide FeatureCard's href to ResourceC…

### LINKED ISSUE #517: FeatureCard has nested anchors
## The FeatureCard component has nested anchors

When implementing the `<FeatureCard>` component, I noticed a warning in the console that says `Warning: validateDOMNesting(...): <a> cannot appear as a descendant of <a>.` 

What browser are you working in?
Chrome, Version 78.0.3904.97 (Official Build) (64-bit)

1. import FeatureCard
2. implement the thing
3. give it an href
4. peep the console 

## Screenshot

anchor_descendent_of_anchor

featurecard_browser_console

:heart: the work y'all do

✊

jendowns:
Closes #517 

As shown in #517, the `FeatureCard` currently takes an `href` and then adds that URL both a wrapper anchor element **and** to the inner `ResourceCard` component, which subsequently renders its own anchor element. As a result, the `FeatureCard` has nested anchors, which is [invalid markup](https://html.spec.whatwg.org/multipage/text-level-semantics.html#the-a-element) & generates a console error.

I propose a couple changes to fix this -- namely:
* the `FeatureCard` should not share its `href` value with the `ResourceCard`, as that is not necessary.
* the `ResourceCard` should render as a `div` if no `href` is provided (similar change to #515)

With these 2 changes, the `FeatureCard` will only render one anchor (the primary wrapper).

#### Changelog

**New**

- add focus styles to `FeatureCard`

**Changed**

- if no `href` is provided to `ResourceCard`, render as a `div`
- `FeatureCard` no longer passes its `href` value down to the inner `ResourceCard`


jendowns:
Tagging @dakahn 👍

packages/gatsby-theme-carbon/src/components/FeatureCard/feature-card.scss
```diff
@@ -33,6 +34,14 @@
   background: $carbon--gray-80; //$hover-ui for gray 90 theme
 }
 
+.#{$prefix}--feature-card__link:focus {
+  outline: none;
+
+  .gatsby-resp-image-wrapper {
+    @include focus-outline;
```
jendowns: Please note that this ends up adding a 1px outline.
But please let me know if you'd like a thicker outline, like 2px. That would be a custom rule outside the mixin (because offset won't work here), but I can switch that easily 👍

vpicone: @jendowns If you could increase it to match the other components that'd be great. We've been using this rather than the focus-outline helper:

```
  outline: 2px solid $focus;
  outline-offset: -2px;
```

jendowns:
Thanks @vpicone -- I just updated the focus outline to be 2px, but I left out an offset rule. Given the element where the outline has to be applied (due to margins etc), a negative offset would not be visible.

PR #533, opened 2019-11-15
fix: homepage template list color

### LINKED ISSUE #532: Lists under `.container--homepage` don't use `color: white` like `.container--dark`
## Detailed description

> Describe in detail the issue you're having.

Using a list on the `Homepage` template results in a black color list on a dark background.

The text color should be white for a list item on a the homepage, which uses `.container--homepage` not `.container--dark`.

Firefox/Chrome

1. Make an index mdx file that uses the homepage template
2. Add a list
3. Go look at the list in browser, it should be "invisible", the background and foreground colors will be the same

## Additional information

image

Text highlighted for visibility.

jonwrona:
Closes #532

<del>Added a style to set the color of a list item on the homepage template</del>

Dark theme on Homepage template works to update the list color.

#### Changelog

**New**

- <del>Added one style in `Markdown.module.scss` to change color of a `.bx--list__item` to white inside a `.container--homepage`</del>
- Added `theme="dark"` prop to Homepage template

**Changed**

- n/a

**Removed**

- n/a


vpicone:
I think instead we should pass `theme="dark"` to the layout component in the homepage. Rather than tacking this style logic onto the home page container.

jonwrona:
@vpicone I went ahead and made that change, makes a lot more sense to do it that way. Thanks for the heads up I wasn't aware that was a prop!

PR #554, opened 2019-11-21
feat(do-dont): adds do dont example row for spacing overrides

### LINKED ISSUE #497: [Spacing] Vertical spacing between image clusters should be 32px
Spacing in between images should be consistent both horizontally and vertically
https://www.carbondesignsystem.com/guidelines/accessibility/overview

Screen Shot 2019-11-05 at 4 43 47 PM

It should behave more like this:
https://gatsby-theme-carbon.now.sh/components/ImageGallery

Screen Shot 2019-11-05 at 4 48 20 PM

FYI 
@vpicone 
@mjabbink

jnm2377:
Closes #497 

`<DoDont>` clusters need vertical spacing (between rows) of `32px` so it's equal to horizontal spacing (between columns) of `32px`. The current row spacing is `24px`. This PR introduces a `<DoDontExampleRow>` component that overrides regular row spacing styles specifically for `<DoDont>` clusters. 

This seemed like the easiest way to do this to allow maximum flexibility in terms of how the columns are laid out inside the row vs. doing it similar to how`<ImageGallery>` & `<ImageGalleryImage>` do it, where `DoDontExample` would receive a `col` prop that renders the image automatically in a column of the specified size. Having props that render the columns automatically could get tedious if folks want to specify column sizes for diff breakpoints. However, am open for feedback on this.

#### Changelog

**New**

- `DoDontExampleRow`

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -1,6 +1,6 @@
 {
   "packages": ["packages/*"],
-  "version": "1.17.1",
+  "version": "1.17.2",
   "npmClient": "yarn",
   "useWorkspaces": true,
   "publish": {
--- a/lerna.json
+++ b/lerna.json
@@ -1,6 +1,6 @@
 {
   "packages": ["packages/*"],
-  "version": "1.17.3",
+  "version": "1.17.2",
   "npmClient": "yarn",
   "useWorkspaces": true,
   "publish": {
--- a/lerna.json
+++ b/lerna.json
@@ -1,6 +1,6 @@
 {
   "packages": ["packages/*"],
-  "version": "1.17.2",
+  "version": "1.17.3",
   "npmClient": "yarn",
   "useWorkspaces": true,
   "publish": {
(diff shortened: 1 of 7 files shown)


vpicone:
I really think the Do/Don't component is in need of a re-write, I despise all the row and column copy and paste. Needs to be using css modules as well.

@jnm2377 What do you think of a `DoDontRow`, a `Do` and a `Dont` component? We can tuck away all of the row/column business and pass through any row/column props in case they want to override the default col or whatever. As an added benefit, we could introduce this new API without breaking old versions, folks could still manually use `DoDontExample` with `Row` and `Column`, we'd maybe just show a console warning with the new api.

I can take this or we can work on it together!

jnm2377:
@vpicone that makes sense. I'm gonna close this PR for now. I'll be out for a few weeks for the holidays, but can definitely work on this in the new year. I'd be happy to tag team if you want.

PR #558, opened 2019-11-22
feat(html): add default lang to site metadata, pass to helmet

### LINKED ISSUE #557: HTML needs default language identified with `lang` attribute
## Detailed description

The `HTML` element does not have a default `lang` set. In the DAP tool, it generates the following violation:

Screen Shot 2019-11-22 at 2 27 16 PM

### What WCAG 2.1 checkpoint does the issue violate?

http://www.ibm.com/able/guidelines/ci162/language_of_page.html

Run the DAP tool on any page using `gatsby-theme-carbon`, and you'll see the this violation.

jendowns:
Closes #557

Adds default language via `lang` attribute to `HTML` element, per [Checkpoint 3.1.1.](https://www.ibm.com/able/guidelines/ci162/language_of_page.html)

#### Changelog

**New**

- add `lang` to `siteMetadata`

### DIFF
--- a/packages/example/gatsby-config.js
+++ b/packages/example/gatsby-config.js
@@ -3,7 +3,6 @@ module.exports = {
     title: 'Gatsby Theme Carbon',
     description: 'A Gatsby theme for the carbon design system',
     keywords: 'gatsby,theme,carbon',
-    lang: 'en',
   },
   pathPrefix: `/gtc`,
   plugins: [
--- a/packages/example/gatsby-config.js
+++ b/packages/example/gatsby-config.js
@@ -3,6 +3,7 @@ module.exports = {
     title: 'Gatsby Theme Carbon',
     description: 'A Gatsby theme for the carbon design system',
     keywords: 'gatsby,theme,carbon',
+    lang: 'en',
   },
   pathPrefix: `/gtc`,
   plugins: [
--- a/packages/gatsby-theme-carbon/src/components/Meta.js
+++ b/packages/gatsby-theme-carbon/src/components/Meta.js
@@ -3,7 +3,7 @@ import { Helmet } from 'react-helmet';
 import { useMetadata } from '../util/hooks';
 
 const Meta = ({ pageTitle, pageDescription, pageKeywords, titleType }) => {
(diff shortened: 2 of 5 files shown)


packages/gatsby-theme-carbon/src/components/Meta.js
```diff
@@ -34,7 +34,9 @@ const Meta = ({ pageTitle, pageDescription, pageKeywords, titleType }) => {
           content: pageKeywords || keywords,
         },
       ]}
-    />
+    >
+      <html lang={lang || 'en'} />
```
jendowns: I think this is nice because even if someone doesn't add `lang` to their own metadata, at least their site will have a valid `lang` value 🤔 But let me know what you think!

vpicone: If they don't add one, it should grab the `en` you have defined in the theme's gatsby-config along with the other defaults (title, keyword etc)

jendowns:
Thanks @vpicone! Just made the change you suggested above 👍

packages/example/gatsby-config.js
```diff
@@ -3,6 +3,7 @@ module.exports = {
     title: 'Gatsby Theme Carbon',
     description: 'A Gatsby theme for the carbon design system',
     keywords: 'gatsby,theme,carbon',
+    lang: 'en',
```
vpicone: I think maybe we should leave this out of the example, it doubles as the starter-kit and folks might not know if they need to leave it in.

Would could update [https://gatsby-theme-carbon.now.sh/guides/configuration#site-title-and-description](this part of the config guide) to just be Site Metadata and add a note there maybe? "**If necessary**, you can change the default language of your document using the `lang` property" (maybe dont need the emphasis haha)

Some thing like 
```suggestion
```

vpicone:
Awesome, thanks @jendowns just had one more quick thing

jendowns:
Thanks @vpicone! I just updated the example/demo site's metadata & also added some info to the configuration guide. Could you take another look when you get a chance?

PR #577, opened 2019-12-02
feat: Add remark plugins option to global configuration, resolves #576

### LINKED ISSUE #576: feat: Add configurability of remark plugins
### Summary

When using gatsby-theme-carbon in my own project I want to be able to extend the list of remark plugins of the theme configuration So I can benefit from additional gatsby remark plugins.

It only requires code changes. No design work is needed.

### Justification

In our development team we want to present visual representation of the product features and architecture in our developers' guide, so that newcomers and existing developers can get a better understanding and smoother onboarding experience. In order to avoid having stale images and outdated visualisation, we want to generate the images from markdown text, so that we can maintain them hand-in-hand with our relevant code changes.

In the future there may be additional markdown plugins that we would be able to add to the configuration if they turn out to be beneficial.

When user want to add for instance graphs and diagrams generated build time from markdown text, she/he would add the following lines to any mdx files:

````
```mermaid
graph LR
install[Install Plugin]
install --> configure[Configure Plugin]
configure --> draw[Draw Fancy Diagrams]
```
````

In case of successful implementation the following image would appear on the built gatsby app:
image

In order to achieve this one should extend the configuration e.g. like this in their own project:

```js
  {
      options: {
        isSearchEnabled: true,
        repository: {
          baseUrl:
            'https://github.com/carbon-design-system/gatsby-theme-carbon',
          subDirectory: '/packages/example',
        },
// extra lines from here:
        gatsbyRemarkPlugins: [ 
          {
            resolve: 'gatsby-remark-mermaid',
            options: {
              language: 'mermaid',
              theme: 'default',
              viewport: {
                width: 200,
                height: 200,
              },
              mermaidOptions: {
                themeCSS: '.node rect { fill: cornflowerblue; }',
              },
            },
          },
        ],
      },
    },
```

See https://www.gatsbyjs.org/packages/gatsby-remark-mermaid/ for further details of this specific example.

### Must-have functionality

- As a theme user I want to extend the Gatsby Remark plugin configuration with new plugins in my own project depending on gatsby-theme-carbon.

### Specific timeline issues / requests

The sooner the better. Without this I can't have living graphs and diagrams in our dev docs (only hardcoded images as a workaround)

### Available extra resources

I'm ready to submit a PR for this.

radnotiz:
Closes #576 

Adds new option 'gatsbyRemarkPlugins' to 'gatsby-theme-carbon' configuration object.

#### Changelog

**New**

- new option with key 'gatsbyRemarkPlugins' with a value type Array

**Changed**

- 'gatsby-theme-carbon' global configuration object

**Removed**

- n/a

### DIFF
--- a/packages/gatsby-theme-carbon/gatsby-config.js
+++ b/packages/gatsby-theme-carbon/gatsby-config.js
@@ -19,8 +19,24 @@ module.exports = themeOptions => {
     lunrOptions = defaultLunrOptions,
     repository,
     mediumAccount = 'carbondesign',
-    gatsbyRemarkPlugins,
+    gatsbyRemarkPlugins = [],
   } = themeOptions;
+
+  const defaultRemarkPlugins = [
+    { resolve: `gatsby-remark-unwrap-images` },
+    { resolve: `gatsby-remark-smartypants` },
+    {
+      resolve: `gatsby-remark-images`,
+      options: {
+        maxWidth: 1152,
+        linkImagesToOriginal: false,
+        quality: imageQuality,
+        withWebp,
+      },
+    },
+    { resolve: `gatsby-remark-responsive-iframe` },
+    { resolve: `gatsby-remark-copy-linked-files` },
+  ];
 
   return {
     siteMetadata: {
@@ -51,22 +67,8 @@ module.exports = themeOptions => {
         options: {
           extensions: mdxExtensions,
           gatsbyRemarkPlugins: [
-            ...[
-              { resolve: `gatsby-remark-unwrap-images` },
-              { resolve: `gatsby-remark-smartypants` },
-              {
-                resolve: `gatsby-remark-images`,
-                options: {
-                  maxWidth: 1152,
-                  linkImagesToOriginal: false,
-                  quality: imageQuality,
-                  withWebp,
-                },
-              },
-              { resolve: `gatsby-remark-responsive-iframe` },
-              { resolve: `gatsby-remark-copy-linked-files` },
-            ],
-            ...(gatsbyRemarkPlugins || []),
+            ...defaultRemarkPlugins,
+            ...gatsbyRemarkPlugins,
           ],
           plugins: ['gatsby-remark-images'],
           remarkPlugins: [remarkSlug],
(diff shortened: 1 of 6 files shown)


packages/gatsby-theme-carbon/gatsby-config.js
```diff
@@ -50,19 +51,22 @@ module.exports = themeOptions => {
...
-            { resolve: `gatsby-remark-unwrap-images` },
-            { resolve: `gatsby-remark-smartypants` },
-            {
-              resolve: `gatsby-remark-images`,
-              options: {
-                maxWidth: 1152,
-                linkImagesToOriginal: false,
-                quality: imageQuality,
-                withWebp,
+            ...[
```
vpicone: I could be wrong, but I think this might be why build is failing. You shouldn't need to spread the array or have the array at all.  Maybe try creating the array above outside the return. See other comment

packages/gatsby-theme-carbon/gatsby-config.js
```diff
@@ -19,6 +19,7 @@ module.exports = themeOptions => {
     lunrOptions = defaultLunrOptions,
     repository,
     mediumAccount = 'carbondesign',
+    gatsbyRemarkPlugins,
```
vpicone: Maybe the default assignment here to follow the other default theme options.

```suggestion
    gatsbyRemarkPlugins = [],
```

radnotiz:
Hi @vpicone, thanks I totally agree with your feedback. I removed the plugin from the demo page, and cleaned up the code based on your suggestions.

packages/gatsby-theme-carbon/gatsby-config.js
```diff
@@ -19,8 +19,25 @@ module.exports = themeOptions => {
     lunrOptions = defaultLunrOptions,
     repository,
     mediumAccount = 'carbondesign',
+    gatsbyRemarkPlugins = [],
   } = themeOptions;
 
+  const defaultRemarkPlugins = [
```
radnotiz: I left it inside the function, because it needs themeOptions.

packages/gatsby-theme-carbon/gatsby-config.js
radnotiz: Done

packages/gatsby-theme-carbon/gatsby-config.js
radnotiz: Done

PR #581, opened 2019-12-03
fix: increase max-height of switcher

### LINKED ISSUE #579: Switcher is too short / IBM iX link cut off
Last link on the switcher is getting cut off, we added a link and didn't account for the additional height.

Screen Shot 2019-12-02 at 7 45 12 PM

*marking as sev1 since the last link (IBM iX) isn't visible/clickable at all

alisonjoseph:
Closes #579

Increases max-height of switcher so that bottom link isn't cut off.

### Review
- make sure all links are visible in switcher on desktop
- @jeanservaas make sure spacing below bottom link is correct, not 100% sure what it should be

### DIFF
--- a/packages/gatsby-theme-carbon/src/components/Switcher/Switcher.module.scss
+++ b/packages/gatsby-theme-carbon/src/components/Switcher/Switcher.module.scss
@@ -36,7 +36,7 @@
     // https://css-tricks.com/using-css-transitions-auto-dimensions/
     // Tied to the height of the app switcher. Allows for animating up to auto.
     // Matching the height exactly is imperitive to keep the correct animation timing.
-    max-height: 492px;
+    max-height: 508px;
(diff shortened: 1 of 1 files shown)


vpicone:
RIP my bullet-proof comment failed me.

packages/gatsby-theme-carbon/src/components/Switcher/Switcher.module.scss
```diff
@@ -36,7 +36,7 @@
     // https://css-tricks.com/using-css-transitions-auto-dimensions/
     // Tied to the height of the app switcher. Allows for animating up to auto.
     // Matching the height exactly is imperitive to keep the correct animation timing.
-    max-height: 458px;
+    max-height: 492px;
```
vpicone: Looks like it's supposed to cut off 24px after the last item per #57 

```suggestion
    max-height: 508px;
```

Looking at a build from 2 weeks ago, it wasn't right even then (too short).

PR #584, opened 2019-12-05
fix: remove container classname scope

### LINKED ISSUE #578: Feature card: Card misaligned on right edge
The white resource card on feature card is not maintaining its position on the edge of image across breakpoints. In XL breakpoint it creates a margin rather than being flush to right side.

This looks like reverted back to something you already fixed in the past @vpicone 

Screen Shot 2019-12-02 at 6 16 16 PM

[Screen Recording 2019-12-02 at 2.51.27 PM.zip](https://github.com/carbon-design-system/gatsby-theme-carbon/files/3913896/Screen.Recording.2019-12-02.at.2.51.27.PM.zip)

Screen Shot 2019-12-02 at 6 16 03 PM

alisonjoseph:
closes #578 

Fix for Carbon homepage, no-gutter classes aren't being applied because the homepage doesn't use the .container class. This should also fix the max breakpoint feature card mis-alignment

#### Changelog

Added the container class back to the homepage template

### Testing

Ensure that this won't break anything - check components using the grid, and the aside component which has nested grid classnames.

### DIFF
--- a/packages/gatsby-theme-carbon/src/components/Container/Container.js
+++ b/packages/gatsby-theme-carbon/src/components/Container/Container.js
@@ -34,7 +34,7 @@ const Container = ({ children, homepage, theme }) => {
   })();
 
   const containerClassNames = classnames({
-    container: theme !== 'dark',
+    container: true,
     'container--homepage': homepage,
     'container--dark': theme === 'dark',
   });
--- a/packages/gatsby-theme-carbon/src/components/Container/Container.js
+++ b/packages/gatsby-theme-carbon/src/components/Container/Container.js
@@ -34,7 +34,7 @@ const Container = ({ children, homepage, theme }) => {
   })();
 
   const containerClassNames = classnames({
-    container: theme !== 'dark' || !homepage,
+    container: theme !== 'dark',
     'container--homepage': homepage,
     'container--dark': theme === 'dark',
   });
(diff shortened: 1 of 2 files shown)


vpicone:
Ahh I think the motivation for `.container` Was to scope these overrides to the MDX content. One way to handle this while controlling the blast radius a bit might be to apply the `container` class in `Container.js` no matter what ( remove the `theme != “dark” || !homepage` check. 

Then the only things impacted by the pr would be homepage and non-mdx pages (404)

alisonjoseph:
I think the homepage must have originally had the container class, but was removed for some reason. Possibly to do with link colors? I'll try adding it back and see what happens.

alisonjoseph:
@vpicone updated to just add the class to the homepage

vpicone:
@alisonjoseph 

The homepage is dark theme so it's still not getting applied:

Screen Shot 2019-12-05 at 11 53 26 PM

Screen Shot 2019-12-05 at 11 56 43 PM

alisonjoseph:
🤦‍♀

alisonjoseph:
ok, updated. Hopefully that doesn't break any other homepage or dark styles.

alisonjoseph:
Tested on Carbon/IDL homepage w/ the container class added and all seems good 🤞

PR #599, opened 2019-12-10
fix(left-nav): add support for nested URLs

SimonFinney:
Currently, the theme doesn't support active navigation item links for URLs with a nested structure - this PR aims to address that.

I don't fully understand the intent behind the original line so perhaps this can be a discussion in order to ensure all cases are covered.

## Expected

Expected

## Actual

Actual

#### Changelog

**Changed**

Check the navigation item link path directly against `pathname`

### DIFF
--- a/packages/gatsby-theme-carbon/src/components/LeftNav/LeftNavItem.js
+++ b/packages/gatsby-theme-carbon/src/components/LeftNav/LeftNavItem.js
@@ -71,7 +71,7 @@ const SubNavItems = ({ items, pathname, onClick }) =>
   items.map((item, i) => {
     const hasActiveTab =
       item.path.split('/').filter(Boolean).length > 2
-        ? pathname.includes(item.path.slice(0, item.path.lastIndexOf('/')))
+        ? item.path === pathname
         : pathname.split('/').toString() === item.path.split('/').toString();
     return (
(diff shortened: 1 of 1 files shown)


vpicone:
@SimonFinney I believe that line is meant to handle pages with tabs. In those cases, the path won't match the link exactly since the link points to the first tab. Example page with tabs: https://www.carbondesignsystem.com/components/accordion/code

SimonFinney:
🤔 Ah, I see it now! We're hoping to migrate our URLs over with little disruption to our current users, but this would most likely force a breaking change and we might need to consider creating redirects for each page.

Would there be another possible solution for identifying tabs over `item.path.split('/').filter(Boolean).length > 2` ?

vpicone:
@SimonFinney not sure :/ would definitely be down for something different though! The nav logic is the most goopy stuff in the theme I think.

jnm2377:
Hi @SimonFinney 👋are you still working on this?

SimonFinney:
@jnm2377 Nope, we've revisited our URL structure and are handling it with client-side redirects instead - this can be closed, unless someone else wants to pick it up

vpicone:
closing in favor of https://github.com/carbon-design-system/gatsby-theme-carbon/pull/698

PR #603, opened 2019-12-11
fix: update carbon and patch list styles

### LINKED ISSUE #559: [lists] List styling
List styles seem very different and don't seem to fit the page layout any more.

## Detailed description
See pics below:
Screen Shot 2019-11-22 at 4 37 32 PM

and in columns:
Screen Shot 2019-11-22 at 4 36 11 PM

When trying to use lists their formatting suddenly appears to be very different from what we have come to expect. Perhaps we have some wrapper that is bonking the styles?

We would like to see the lists formatted as they were before when we were using 1.14.3

Chrome primarily.

1. Use a typical markdown list
2. Be using the latest GTC 1.17.2
3. Spool up the dev environment or do a build
4. View the page with a list in browser

vpicone:
closes #559

### DIFF
--- a/packages/gatsby-theme-carbon/src/components/Header/Header.js
+++ b/packages/gatsby-theme-carbon/src/components/Header/Header.js
@@ -6,7 +6,7 @@ import {
   SkipToContent,
   HeaderGlobalBar,
   HeaderGlobalAction,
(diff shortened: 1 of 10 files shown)


alisonjoseph:
Looks like we're missing the nested list bullet?

PR #607, opened 2019-12-11
fix(skip-to-content): add explicit href to component, add id to…

### LINKED ISSUE #606: [SkipToContent]: does not target any part of the page to allow skipping
## Detailed description

The `SkipToContent` component is meant to link out to the main section of a page, to allow keyboard users to skip directly to that content.

Unfortunately, while the `SkipToContent` component currently links to `#main-content`, there is currently no element/target with that id on any of the pages. As a result, the `SkipToContent` link does not go anywhere. 

Ideally, there would be an element on the page somewhere with the `id="main-content"` so that when a user activates the `SkipToContent` link, they can jump to that element/target.

jendowns:
Closes #606

The `SkipToContent` component needs to target part of the page so that keyboard users can jump to it, so I suggest adding `id="main-content"` to the `<main>` element of each page. NOTE: this would *not* include homepage, since it doesn't appear to have a `SkipToContent` link -- is that acceptable?

Also, since the default `href` for the `SkipToContent` is `#main-content`, I suggest explicitly defining it where the component is used so that folks who come along later can quickly see where the link goes. 👍 

#### Changelog

**New**

- add `href="#main-content"` to the `SkipToContent` component
- add `id="main-content"` to the `<main>` element on each (non-homepage) page.

### DIFF
--- a/packages/gatsby-theme-carbon/src/components/Container/Container.js
+++ b/packages/gatsby-theme-carbon/src/components/Container/Container.js
@@ -48,7 +48,11 @@ const Container = ({ children, homepage, theme }) => {
         role="presentation"
         tabIndex="-1"
       />
-      <main aria-hidden={overlayVisible} className={containerClassNames}>
+      <main
+        id="main-content"
+        aria-hidden={overlayVisible}
+        className={containerClassNames}
+      >
         {children}
       </main>
     </>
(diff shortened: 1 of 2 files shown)


vpicone:
I like explicitly defining it as well, thanks for this!

PR #608, opened 2019-12-12
fix: wire up main content to main element

### LINKED ISSUE #606: [SkipToContent]: does not target any part of the page to allow skipping

vpicone:
closes #606


vpicone:
RIP duplicate of #607

PR #611, opened 2019-12-14
Tabs refactor

vpicone:
Refactor of the gatsby tabs component to use IDL design in place of the carbon tabs component.

Testing: implements wcag recommended keyboard navigation: https://www.w3.org/TR/wai-aria-practices/examples/tabs/tabs-1/tabs.html

Need a mobile spec @jeanservaas @mjabbink. The IDL website has short enough tab titles that they don't have to worry about overflow.


packages/gatsby-theme-carbon/src/components/Tabs/Tabs.module.scss
```diff
@@ -1,106 +1,57 @@
...
-  }
+.tab-list {
+  display: flex;
+  margin-left: -1rem;
+  margin-right: -1rem;
+  scrollbar-width: none;
+}
+
+.tab-list li:focus-within {
+  outline: 2px solid blue;
```
alisonjoseph: Would we want to use our $focus color token here? Focus states look like they need a design spec also

vpicone: @jeanservaas is takin a look at if we need a more formal spec for some aspects

mjabbink:
The tab height should be 48px

In mobile goes to drop down

See IDL in desktop and mobile here:
https://www.ibm.com/design/language/typography/typeface#sub-families

**Desktop**
Screen Shot 2019-12-17 at 6 40 37 PM

**Mobil:**
Screen Shot 2019-12-17 at 6 40 47 PM

mjabbink:
Also looks like in Chrome mobile there is 2 or 3 pixels between tabs


File (7)

vpicone:
@mjabbink make sure you're grabbing the latest version, the awkward space was from an old build. This should be the latest: https://gatsby-theme-carbon-git-fork-vpicone-tabs-refactor.carbon-design-system.now.sh/components/Tabs

new:
- added border bottom to tabs list
- use gray-20 for border color
- switch to dropdown on mobile

alisonjoseph:
This is looking great! 

Couple questions
- should the cursor change on hover (it does on the Carbon tabs)
- Mobile tabs - is the dropdown accessible? I believe there was talk of removing it in core Carbon because it wasn't.

vpicone:
@alisonjoseph the Carbon version uses links instead of buttons as the base so that's why they have pointer cursors. But then again so do our `Button` components... This article changed my view on it: https://medium.com/simple-human/buttons-shouldnt-have-a-hand-cursor-b11e99ca374b

As another data point: the WCAG tab component example where they don't use a pointer https://www.w3.org/TR/wai-aria-practices/examples/tabs/tabs-1/tabs.html

For mobile, this solution is accessible because we use React to selectively render _either_ dropdown _or_ tabs so we don't have to try to manage what's hidden/tabbable with css. So as long as they're both independently accessible, the overall component is as well.

vpicone:
@alisonjoseph yeah I didn't realize Carbon gives pointers to buttons by default so doesn't really make sense to diverge here.

mjabbink:
I was indeed looking at the wrong preview link. So looking good now. I think the only remaining detail is the dropdown height in mobile. I think it should also be 48px no? @jeanservaas

mjabbink:
see typetester in desktop and mobile
https://www.ibm.com/design/language/typography/typeface#type-tester

vpicone:
@mjabbink We added 48px 'xl' dropdown in the upcoming Carbon release, I'll be sure to use it.

mjabbink:
one thing hitting me is that the tabs used for the top of the page follow a different width rule. I think those are nice. Will this refactor effect those? I think it’s worth maintaining that rule and having the less wide approach for inside the page.

see this example of nav tabs

Screen Shot 2019-12-19 at 6 27 57 PM

vpicone:
@mjabbink no, those are a separate component so won’t be affected.

mjabbink:
Perfect 
👍

vpicone:
@jeanservaas update the focus outline. Since the only tab that can be focused is the selected one, I used a white hairline to match the button.

PR #617, opened 2019-12-16
Caption spacing fix

### LINKED ISSUE #589: Caption example: theme site
the caption example has too much space above it in the example which should be corrected.
https://gatsby-theme-carbon.now.sh/components/Caption#example. Our own visual example of the usage has to be correct.

Screen Shot 2019-12-09 at 8 03 42 AM

vpicone:
Closes #589

### DIFF
--- a/packages/example/src/pages/components/Caption.mdx
+++ b/packages/example/src/pages/components/Caption.mdx
@@ -13,7 +13,7 @@ The `<Caption>` component is typically used below images or videos. They will de
 
 #### Normal
 
-<Video src="/videos/hero-video.mp4" poster="/images/poster.png" />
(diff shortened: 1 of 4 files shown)


vpicone:
@alisonjoseph that video had an aspect ratio wrapper that made it look like the caption had more space than it did. I updated it to match the do/dont example space now.

alisonjoseph:
I saw you removed the video example, will the caption display correctly underneath all the things? (videos, svgs, images)

vpicone:
@alisonjoseph videos recede from their bottom edge a bit to preserve aspect ratio so at some viewport widths the space varries but there's not much we can do about that.

PR #661, opened 2020-01-16
fix(homepage): use blue40 for links on dark homepage container

jendowns:
Resolves https://github.com/carbon-design-system/carbon-website/issues/792
Resolves https://github.com/carbon-design-system/carbon-website/issues/536

The links on the dark container of the homepage are not styled for `:hover`, `:active`, `:active:visited`, or `:active:visited:hover` states. As a result, these states show default `bx--link` colors which are not accessible against the dark background. (The also applied to the focus outline).

#### Changelog

**New**

- add example link outside of homepage callout so that the modified link styles can be easily tested.

**Changed**

- add style overrides for links on the dark container background of the homepage.

### DIFF
--- a/packages/gatsby-theme-carbon/src/components/Homepage/homepage.scss
+++ b/packages/gatsby-theme-carbon/src/components/Homepage/homepage.scss
@@ -8,17 +8,11 @@
 .container--dark a.#{$prefix}--link:hover,
 .container--dark a.#{$prefix}--link:active,
 .container--dark a.#{$prefix}--link:active:visited,
-.container--dark a.#{$prefix}--link:active:visited:hover,
-.container--homepage a.#{$prefix}--link,
-.container--homepage a.#{$prefix}--link:hover,
-.container--homepage a.#{$prefix}--link:active,
-.container--homepage a.#{$prefix}--link:active:visited,
-.container--homepage a.#{$prefix}--link:active:visited:hover {
+.container--dark a.#{$prefix}--link:active:visited:hover {
   color: $carbon--blue-40;
 }
 
-.container--dark a:focus,
-.container--homepage a:focus {
+.container--dark a:focus {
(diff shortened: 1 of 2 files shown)


packages/gatsby-theme-carbon/src/components/Homepage/homepage.scss
```diff
@@ -4,11 +4,24 @@
...
+.container--dark a.#{$prefix}--link,
+.container--dark a.#{$prefix}--link:hover,
+.container--dark a.#{$prefix}--link:active,
+.container--dark a.#{$prefix}--link:active:visited,
+.container--dark a.#{$prefix}--link:active:visited:hover,
+.container--homepage a.#{$prefix}--link,
+.container--homepage a.#{$prefix}--link:hover,
+.container--homepage a.#{$prefix}--link:active,
+.container--homepage a.#{$prefix}--link:active:visited,
+.container--homepage a.#{$prefix}--link:active:visited:hover {
```
jendowns: @vpicone Can I just remove this rules for `container--homepage`? Based on what I've seen here and on the Carbon website, it's the `container--dark` class that actually matters here.

jendowns: Additional note: these selectors need this level of specificity due to the rules they are overriding: .[classname].[classname]:hover, etc
But let me know if you have concerns 👍

vpicone: @jendowns yeah you got it exactly right. The homepage as a dark theme by default so that's the one that matters. Thanks for taking the time to do this!

jendowns: Thanks @vpicone! I pushed an update for this

PR #674, opened 2020-01-19
fix(callout-component): add padding-top to p of first column

### LINKED ISSUE #664: fix[callout component]: H3 sitting too high; needs to break to two lines
## fix[callout component]: H3 sitting too high; needs to break to two lines

- [ ] Move H3 in Gatsby Callout component down 3px so it optically aligns with the larger paragraph to the right.

- [ ] H3 in Callout component needs to break to two lines

xxxle0:
Closes #664 

i add padding top to p of first column

_wait for confirm reduce width of first column._


vpicone:
Should we only be doing this at large+ breakpoints?

jnm2377:
> Should we only be doing this at large+ breakpoints?

@vpicone I think it would be starting at medium breakpoint. Looks like it's still horizontal up until `672px`

Screen Shot 2020-01-20 at 11 09 24 AM

Screen Shot 2020-01-20 at 11 10 12 AM


@xxxle0 to answer your question though, I don't think that reducing the column is what we want however, since that will change the alignment of the second column.

vpicone:
@jnm2377 ah yeah good point.

jnm2377:
I haven't come across an easy way to manipulate line breakpoints other than hardcoding `br`/`wbr`, wrapping the breakpoint in a `span` and giving that different styles, or giving the container some right padding. I've looked through several stack overflow and css-tricks articles and that's about all I've been able to come up with. Nothing truly allows us to control where a line breaks. I think the only thing we could possibly control on our end would be the right padding, since we can't foresee what the text will be for people. Going forward, I think we can either say that individual sites can hardcode a line break, or we can add right padding. What are your thoughts @vpicone ?

vpicone:
@jnm2377 I think you're right, we'll never be able to meet everyone's needs with trying to force it with padding-right. It should be controlled with site-supplied breaks so they can fineness they're exact needs.

PR #676, opened 2020-01-21
feat(GifPlayer): adds component to pause and play gifs for a11y 

jnm2377:
Rel https://github.com/carbon-design-system/design-language-website/issues/411

This component adds pause and play functionality for gifs by replacing the gif with a static image (much like the way `ArtDirection` uses different images).

It made the most sense to add this component to Gatsby since we use gif's in our Carbon site and IDL site.

#### New

- `GifPlayer`

#### Changed

- `ImageGallery` functionality to work with `GifPlayer`

#### Review
- Go to the `GifPlayer` page in the left nav to see a working example
- Also review the example in the `ImageGallery` page
- Check for visually correct styles
- Check for a11y issues (I ended up adding `aria-hidden="true/false"` bc even though the styles use `display:none` to toggle between the gif/static image, I still see the hidden images in the DOM). cc: @dakahn


packages/gatsby-theme-carbon/src/components/GifPlayer/GifPlayer.js
```diff
@@ -0,0 +1,101 @@
...
+      </div>
+      <div
+        className={staticImageClassNames}
+        aria-hidden={paused ? false : 'true'}
+      >
+        {childrenArray[1]}
+      </div>
+      <button
+        type="button"
+        aria-label="play pause toggle"
```
dakahn:
```suggestion
        aria-label="animation play pause toggle"
```
Just a hair more specific here and I think our intent is way clearer 👍

packages/gatsby-theme-carbon/src/components/GifPlayer/GifPlayer.js
```diff
@@ -0,0 +1,101 @@
...
+        {childrenArray[0]}
+      </div>
+      <div
+        className={staticImageClassNames}
+        aria-hidden={paused ? false : 'true'}
+      >
+        {childrenArray[1]}
+      </div>
+      <button
+        type="button"
```
dakahn:
```suggestion
```

dakahn: Don't think we need this here. `<button type="button">` as far as I know communicates that the button does nothing when pressed.

[reference](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button)

packages/gatsby-theme-carbon/src/components/GifPlayer/GifPlayer.js
```diff
@@ -0,0 +1,101 @@
...
+  const childrenArray = React.Children.toArray(children);
+
+  return (
+    <div className={containerClassNames}>
+      <div className={gifClassNames} aria-hidden={paused ? 'true' : false}>
+        {childrenArray[0]}
+      </div>
+      <div
+        className={staticImageClassNames}
+        aria-hidden={paused ? false : 'true'}
```
dakahn:
```suggestion
```

I think I had something else in my mind when I suggested this -- sorry. It's totally cool if the gif is still in the DOM when it's paused.

packages/gatsby-theme-carbon/src/components/GifPlayer/GifPlayer.js
```diff
@@ -0,0 +1,101 @@
...
+      >
+        {childrenArray[1]}
+      </div>
+      <button
```
dakahn:
```suggestion
      <button
      aria-pressed={paused ? 'true' : 'false'}
```

This will communicate that the button is a toggle (a little redundant given our label, but expected. And can also let the screen reader user know if the button is toggled on or off potentially.

dakahn:
As an aside -- it could be cool to explore making the gif player an `aria-live` region that can notify the user that the gif has been played or paused. We use this pattern in our Loading component:
2020-01-22 17_07_39-Window

packages/gatsby-theme-carbon/src/components/GifPlayer/GifPlayer.js
jnm2377: @dakahn I added this because the gatsby theme eslint rules require a `type` to be specified for buttons, and I wasn't sure if this would be considered `type="submit"` if it's not submitting any data. cc: @vpicone

vpicone: @dakahn @jnm2377 type button is correct buttons default to type submit so you need to define them as type button to avoid awkward form submission behavior. 
Type=“button” is what you’d expect to see here.

dakahn: Reading the link I referenced

_The button **has no default behavior** and does nothing when pressed. It can have client-side scripts associated with the element's events, which are triggered when the events occur._

So I think you're right -- but the only part that's throwing me is the "does nothing when pressed" bit. I'm not sure if that would be communicated to screen reader users who we would want to expect the button in question to open a menu. 

Not a blocker tho!

jnm2377: @dakahn not sure if it makes a difference, but when I test this with voice over it says it's a toggle button. Is that not a sufficient indication of what it does? I also updated the button label to say if it's playing or paused.

jnm2377: I'm not sure what is the expected voice over but this is what it does right now. 
voiceover

dakahn: That's great!

PR #685, opened 2020-01-27
Remove target _blank in resource card and feature card

### LINKED ISSUE #678: Remove target="_blank" from Theme
All links created using markdown stay on the same tab, which is the default behavior.

The `ResourceCard` and `FeatureCard` components both need to match this. They currently open in a new tab.

image

Users of the theme should be able to pass `target="_blank"` as a prop if they need it.

xxxle0:
Closes #678 

Remove target _blank in ResourceCard and FeatureCard


xxxle0:
please review and don't merge it. I need confirmation for ArticleCard and ImageCard

PR #704, opened 2020-02-04
feat: add bundle analyzer

vpicone:
test: 
- run `build:analyze`
- check out `/guides/npm-scripts` for new content


packages/example/src/data/nav-items.yaml
```diff
@@ -22,6 +22,8 @@
       path: /guides/styling
     - title: Navigation
       path: /guides/navigation/sidebar
+    - title: Navigation
```
sstrubberg: rename to scripts or NPM scripts or something like that..
image

vpicone: I hate how they don't capitalize 😢

sstrubberg:
after i run `yarn build:analyze`, it runs, but then it just stops. 127.0.0.1:3001 is just a "Page can not be displayed"

vpicone:
@sstrubberg should be good now

PR #731, opened 2020-02-19
fix: use grid for next-prev

### LINKED ISSUE #730: Next/Prev buttons alignment
The next buttons in footer are out of sync.

Screen Shot 2019-12-17 at 2 00 00 PM

vpicone:
closes #730 

new: uses grid components for next/prev


jnm2377:
I think the issue was actually referring to the buttons not having the correct "next/previous" link. Not necessarily the grid alignment. Not to say there's anything wrong with this PR. But I think we may have to re-open the issue, and it seems like maybe it would be IDL specific. 🤷‍♀

PR #745, opened 2020-02-26
fix(Lists): Adjusted margins

### LINKED ISSUE #720: Ordered list is not aligned to the grid
This screenshot shows how the ordered list are rendered outside of the grid.

May need to override `left` on `.bx--list--ordered:not(.bx--list--nested) > .bx--list__item::before` to be `-1rem`?

image

sstrubberg:
Closes #720 

I don't know where all those commits came from 😬


packages/gatsby-theme-carbon/src/styles/internal/_page.scss
```diff
@@ -166,3 +166,21 @@ img {
...
+.bx--list--ordered {
+  margin-left: 1rem;
+}
+
+//---------------------------------------
+// Nested bullet alignment
+//---------------------------------------
+
+.bx--list--nested .bx--list--unordered,
+.bx--list--ordered {
```
jnm2377: I don't think these selectors are doing anything. When I inspect it, the margin is still 1rem. 
`bx--list--unordered` is the same as `bx--list--nested`, as in they're both the `ul` or `ol`. Seems like it should either be:

```css
.bx--list--unordered.bx--list--nested, 
.bx--list--ordered.bx--list--nested { 
...
}
```

 or

```css
.bx--list--nested .bx--list__item {
...
}
```
if you're trying to target the item inside.

PR #750, opened 2020-03-03
fix: code overflow closes #749

### LINKED ISSUE #749: [Code]: Code component url overflows on mobile causing jank.
IMG_0812

vpicone:


vpicone:
@sstrubberg no one's implemented it yet, but `text-overflow: fade` would be really dope: https://developer.mozilla.org/en-US/docs/Web/CSS/text-overflow#fade

vpicone:
Talking with @jeanservaas we're going to go with the same fade as the code below.

vpicone:
closes #763

vpicone:
Fade out my take a bit to get to, merging this as a stop gap.

PR #753, opened 2020-03-03
fix: inline code font size

### LINKED ISSUE #459: fix[inline code snippet]: sizing incorrect
The inline code snippet sizing has gone wonk: 

Originally we had a 14px inline code snippet to be used with body copy and a 12px inline code snippet to be used in tables 

Looks like now the main inline code snippet has is the 12px and the inline code snippet in the tables is TINY... it's calling 10.5px type which doesn't exist in our scale.

**body inline snippet (is 12px should be 14px):**

image

**table inline snippet (is 10.5px should be 12px):**

image

alisonjoseph:
Closes #459

### DIFF
--- a/packages/gatsby-theme-carbon/src/components/Code/Code.module.scss
+++ b/packages/gatsby-theme-carbon/src/components/Code/Code.module.scss
@@ -18,6 +18,10 @@
     color: $text-01;
   }
 
+  td > code {
+    font-size: 12px;
+  }
+
   // Add's additional length for scrolling under sidebar
   .token:last-of-type {
     margin-right: 64px;
--- a/packages/gatsby-theme-carbon/src/components/Code/Code.module.scss
+++ b/packages/gatsby-theme-carbon/src/components/Code/Code.module.scss
@@ -11,7 +11,7 @@
     position: relative;
     display: inline;
     padding: 0 0.5em;
-    font-size: 0.75em;
+    font-size: 0.875em;
     bottom: 0.0625em;
     border-radius: 4px;
     background-color: $ui-03;


packages/gatsby-theme-carbon/src/components/Code/Code.module.scss
```diff
@@ -11,13 +11,17 @@
...
     bottom: 0.0625em;
     border-radius: 4px;
     background-color: $ui-03;
     color: $text-01;
   }
 
+  td > code {
+    font-size: 12px;
+  }
+
```
alisonjoseph: Without this it was computing to 12.25px instead of 12px. Has something to do with the specificity in the page-table scss file.

packages/gatsby-theme-carbon/src/components/Code/Code.module.scss
```diff
@@ -11,13 +11,17 @@
     position: relative;
     display: inline;
     padding: 0 0.5em;
-    font-size: 0.75em;
+    font-size: 0.875em;
```
vpicone:
```suggestion
    font-size: 0.75rem;
```

vpicone: looks like it shoulda been `rem` so it's relative to the root instead of whatever the current font-size is.

alisonjoseph: @vpicone changing it to `.75rem` makes it too small(12). should be 14px for body text.

alisonjoseph:
@vpicone just updated to `.875rem` for body text and `.75rem` for text inside table cells. Looks like both are rendering correctly now at 14px / 12px.

packages/gatsby-theme-carbon/src/components/Code/Code.module.scss
vpicone: Whoops! My bad, I’d just leave the r in there so it’s /16 instead of /whatever the current font size is.

packages/gatsby-theme-carbon/src/components/Code/Code.module.scss
vpicone: I think It was computing to 12.25 because it was using em. So when the font size goes down to 14, .875*14 = 12.25

PR #755, opened 2020-03-04
fix(NextPrev): offset and hompage conditions WIP

### LINKED ISSUE #746: fix(homepage):NextPrev
image

The first page after the homepage should be in the space to the right.

sstrubberg:
Closes #746 

Added logic and style for the first button to be hidden if it's the homepage, thus giving the offset.

Blocker: It thinks every page is the homepage, displaying the next, but hiding the previous.


vpicone:
@sstrubberg you’ll The way you have isHomepage there is just saying isHompage=true. If you just write the prop name it counts as true. You gotta pass a value in : isHomepage={isHomepage}

PR #765, opened 2020-03-05
added logic and style

sstrubberg:
Closes #

{{short description}}

#### Changelog

**New**

- {{new thing}}

**Changed**

- {{change thing}}

**Removed**

- {{removed thing}}

### DIFF
--- a/packages/gatsby-theme-carbon/src/components/NextPrevious/NextPrevious.js
+++ b/packages/gatsby-theme-carbon/src/components/NextPrevious/NextPrevious.js
@@ -144,12 +144,10 @@ const NextPreviousContainer = props => {
       };
     }
 
-    return currentTitle === 'Home'
-      ? {}
-      : {
-          to: '/',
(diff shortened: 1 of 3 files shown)


packages/gatsby-theme-carbon/src/components/NextPrevious/NextPrevious.js
```diff
@@ -189,7 +187,11 @@ const NextPreviousContainer = props => {
   }
 
   return (
-    <NextPrevious previousItem={getPreviousItem()} nextItem={getNextItem()} />
+    <NextPrevious
+      isHomepage
```
vpicone: You'll need to grab `isHomepage` from the props when we destructure them at line 102

```suggestion
      isHomepage={isHomepage}
```

The part where you pass in just `isHomepage` or `isHomepage={true}` is when you use the NextPrevious component in the `src/templates/Homepage` template

PR #789, opened 2020-03-18
chore: add carbon-icons peer

vpicone:


vpicone:
@sstrubberg it's a legacy dep of carbon-components-react

PR #819, opened 2020-04-21
fix: left nav duplicate active selection and jumping scroll

### LINKED ISSUE #345: QA: nav jumps back up to top when you click on a page below "fold"
Nav should stay in place when new link is selected. Probably only a problem with components.
feb-08-2019 11-18-08

### LINKED ISSUE #386: LeftNav/Tabs active selection bugs
Continuing problems we're seeing with the left nav selection states:

#### 1. Duplicate selection when page names are similar (similar to #44)
  - image
</details>

#### 2. Single pages with tabs lose left nav selection when a tab non-first tab is selected (similar to #168)
  - > First tab selected, has appropriate active state:
image 
  - > Second tab selected, _does not_ have appropriate active state:
image

jnm2377:
Closes #386 
Closes #345


packages/gatsby-theme-carbon/src/util/hooks/useNavScroll.js
```diff
@@ -0,0 +1,16 @@
+import { useEffect, useContext } from 'react';
+import NavContext from '../context/NavContext';
+
+const useNavScroll = () => {
+  const { updateNavScrollOffset } = useContext(NavContext);
+
+  useEffect(() => {
+    const sideNav = document.querySelector('ul.bx--side-nav__items');
```
vpicone, line 8: Could we instead have the hook take a ref? in general doing query selectors in react can get hairy really quick.

packages/gatsby-theme-carbon/src/util/context/NavContext.js
```diff
@@ -13,24 +14,31 @@ const reducer = (state, action) => {
...
+    case 'update-nav-scroll':
+      return { scrollOffset: action.payload };
     default:
       return { ...state, [action.nav]: !state[action.nav] };
   }
 };
 export const NavContextProvider = ({ children }) => {
   const [
-    { leftNavIsOpen, searchIsOpen, switcherIsOpen },
+    { leftNavIsOpen, searchIsOpen, switcherIsOpen, scrollOffset },
```
vpicone, line 25: could we call this something more descriptive leftNavScrollOffset?

vpicone:
Lookin good! do you mind doing a quick video of how we can test this/see it in action?

jnm2377:
Here's an example with a duplicate accordion menu item in different menus, and you can see they're not both highlighted when the accordion page is active. 

leftnav

sstrubberg:
image

:(

sstrubberg:
image

It's killing both Chrome and FF when after I click on something in the leftnav

jnm2377:
@sstrubberg hmmm... I wonder if it's bc of the calculated scrolling that I added so that the left nav doesn't jump. :/ 

Is the duplicate active nav bug still present too?

sstrubberg:
sorry, edited some of your code because the linter was yelling at me. That caused the stalling bug. When i ran it with just your code, both browsers work. However, i'm still getting the original issue of selecting both items.

image

jnm2377:
@sstrubberg I just pushed what I'm testing with. The duplicate accordion nav items. Is that still showing the issue for you?

sstrubberg:
Huzzah! not sure what was different, but both the deploy preview and my local environment are singing. 

However, that linting error is still there.
image

sstrubberg:
Should we just give it an `// eslint-disable-line react-hooks/exhaustive-deps`?

vpicone:
@sstrubberg @jnm2377 those eslint react hooks rules from the react team are super well thought out, I'm down for disabling some rules for old code but we it looks like they've caught something that's a little off here.

vpicone:
Since the left nav is the only thing that cares about the scroll position, I'm not sure it makes sense to lift it all the way up to the app level context. Gatsby let's you pass `state` to a link, maybe we pass that into our side nav items and use it via `location.state` to set the position after the page loads. https://www.gatsbyjs.org/docs/gatsby-link/#pass-state-as-props-to-the-linked-page

vpicone:
I think it's looking great. Only issue now is when there's two menu's open and one closes after navigation, the new scroll position is off. We can either keep track of the change in scrollHeight or update the left nav to persist the open state of each nav menu.

jnm2377:
@vpicone @sstrubberg this is ready for re-review.

PR #821, opened 2020-04-22
feat: add remarkPlugins theme option

alaaalii:
Add a [`remarkPlugins`](https://www.gatsbyjs.org/packages/gatsby-plugin-mdx/#remark-plugins) option to pass along to `gatsby-plugin-mdx`.

The reason this is required in addition to the `gatsbyRemarkPlugins` option is because `gatsbyRemarkPlugins` has had numerous issues with getting plugins to work with it:
- https://github.com/gatsbyjs/gatsby/issues/21866
- https://github.com/gatsbyjs/gatsby/issues/16983
- https://github.com/gatsbyjs/gatsby/issues/8009
- https://github.com/gatsbyjs/gatsby/issues/9103
- https://github.com/gatsbyjs/gatsby/issues/6648

But the `remarkPlugins` option works great and is usually the workaround that is eventually used. So exposing this option to the user helps.

#### Changelog

**New**

- added a `remarkPlugins` theme option to be passed to `gatsby-plugin-mdx`
- added an example on how to use it with the `remark-grid-tables` remark plugin

### DIFF
--- a/packages/example/src/pages/guides/configuration.mdx
+++ b/packages/example/src/pages/guides/configuration.mdx
@@ -203,6 +203,7 @@ plugins: [
 ## Additional remark plugins and override existing remark plugin configuration
 
 - `gatsbyRemarkPlugins` - An array containing gatsby remark plugin configurations to be added/overridden.
+- `remarkPlugins` - An array containing [remark plugin configurations](https://www.gatsbyjs.org/packages/gatsby-plugin-mdx/#remark-plugins) to be added.
 
 ### Example [Gatsby Remark Mermaid plugin](https://www.gatsbyjs.org/packages/gatsby-remark-mermaid/)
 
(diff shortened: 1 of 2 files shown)


packages/gatsby-theme-carbon/gatsby-config.js
```diff
@@ -90,6 +91,9 @@ module.exports = (themeOptions) => {
             ...defaultRemarkPlugins,
             ...gatsbyRemarkPlugins,
           ],
+          remarkPlugins: [
+            ...remarkPlugins,
+          ],
```
vpicone, line 96: I think we can just pass this through for now since we're not adding any defaults of our own.
```suggestion
          remarkPlugins,
```

packages/example/src/pages/guides/configuration.mdx
```diff
@@ -250,3 +251,38 @@ to turn into an image one should add the following configuration to her/his own
...
++-------+----------+------+
+| multi | cells can be    |
+| line  | *formatted*     |
+|       | **paragraphs**  |
+| cells |                 |
+| too   |                 |
++-------+-----------------+
+```
+
+to turn into the table it describes one should add the following configuration to her/his own project:
```
vpicone, line 277:
```suggestion
After installing the `remark-grid-tables` plugin, add it to the `remarkPlugins` array in the theme options.
```

packages/gatsby-theme-carbon/gatsby-config.js
alaaalii: Done!

packages/example/src/pages/guides/configuration.mdx
alaaalii: Done!

alaaalii:
Thanks. I added this to the documentation:

>The `gatsbyRemarkPlugins` option is compatible with all gatsby-remark-* plugins, for example `gatsby-remark-mermaid`. If a gatsby remark plugin does not work, you can try calling the (underlying) remark plugin directly using the `remarkPlugins` option.

Hopefully this wording is good enough to explain the difference?

vpicone:
@alaaalii This is available under `gatsby-theme-carbon@next` we'll generally publish at the end of the week.

PR #826, opened 2020-04-27
feat(ResourceCard): Added Mini

### LINKED ISSUE #794: [Gatsby] Component Request: Add mini resource card to Theme

sample here:

sstrubberg:
Closes #794 

This component adds a `type` prop to the ResourceCard component. If the word `mini` is passed, then it will give you the new Mini ResourceCard, which has everything from the original ResourceCard gutted out except for the title and the children. The children are the icons that get passed. If no icons are passed as children, a default actionIcon of `arrowRight` will be given.

image

For dev reviewers. I think I need to re-work the icons so i'm not repeating myself. Definitely open to feedback. I also made this crazy prop called `noGutterSmYesGutterLg`, which might be the most ridiculous thing ever, but i couldn't figure how to do the negative margins on the Columns for small and medium screens, then drop the negative margins at large screens.

I feel like I got the docs about 80% there. I would love some help to tighten them up.


jeanservaas:
looking great couple of things

- they just need that gray 20 border between them that the big resource cards have it's the $ui-03 subtle border.

- type and icon look like they're riding a little bit high of vertical center, like maybe a pixel

image

sstrubberg:
Ok @jeanservaas. I think I addressed all your feedback. Let me know what you think.

packages/gatsby-theme-carbon/src/components/Grid.js
```diff
@@ -49,6 +49,7 @@ export const Column = ({
   noGutterLg,
   noGutterXl,
   noGutterMax,
+  noGutterSmYesGutterLg,
```
vpicone, line 52: hmm, this seems a little spicy. Does we could have thousands of combinations if this pattern continued.

I assume adding the `no-gutter-sm` impact the larger breakpoints?

packages/gatsby-theme-carbon/src/components/ResourceCard/ResourceCard.js
```diff
@@ -13,100 +13,140 @@ import { settings } from 'carbon-components';
...
+    [`${prefix}--resource-card--disabled`]: disabled,
+    [`${prefix}--resource-card--dark`]: color === 'dark',
+  });
+
+  const aspectRatioClassNames = classnames([`${prefix}--aspect-ratio`], {
+    [`${prefix}--aspect-ratio--2x1`]: aspectRatio === '2:1',
+    [`${prefix}--aspect-ratio--1x1`]: aspectRatio === '1:1',
+    [`${prefix}--aspect-ratio--16x9`]: aspectRatio === '16:9',
+    [`${prefix}--aspect-ratio--4x3`]: aspectRatio === '4:3',
+  });
```
vpicone, line 45: The arguments to `classnames()` don't have to be an array, I think we can also use a string template to clean this up and prevent a never ending list scenario.

```suggestion
	const aspectRatioClassNames = classnames(
      `${prefix}--aspect-ratio`,
      `${prefix}--aspect-ratio--${aspectRatio}`
    );
```

Maybe double check some of the other classnames function for the unnecessary arrays

packages/gatsby-theme-carbon/src/components/ResourceCard/ResourceCard.js
```diff
@@ -13,100 +13,140 @@ import { settings } from 'carbon-components';
...
+    cardContainer = (
+      <a href={href} className={carbonTileclassNames} {...rest}>
+        {cardContent}
+      </a>
     );
   }
-}
+
+  return (
+    <>
```
vpicone, line 135: A lot of these fragments are unnecessary, see which one's you can remove and get away with it.

You'll should only need them very rarely, when you need two adjacent components that don't make sense to have wrapped in some container.

packages/gatsby-theme-carbon/src/styles/internal/_page.scss
```diff
@@ -134,6 +134,15 @@ img {
   padding-right: 0;
 }
 
+.container .#{$prefix}--no-gutter-sm--yes-gutter-lg {
```
vpicone, line 137: Adding rules to the _page should be avoided at all costs, this stylesheet has in the past become a sort of dumping ground for global rules. 

Like I mentioned in the earlier comment you can imagine things getting spicy if we keep trying to put behavior like this into sass.

e.g. `.#{$prefix}--no-gutter-sm--yes--gutter-lg--no-gutter-max`

packages/gatsby-theme-carbon/src/components/ResourceCard/ResourceCard.js
sstrubberg: When I remove them, the page won't won't render. Do i just use divs instead?

packages/gatsby-theme-carbon/src/components/ResourceCard/ResourceCard.js
sstrubberg: Got it. When I do this, users will need to pass `2x1` instead of `2:1`. Does that make a breaking change?

packages/gatsby-theme-carbon/src/components/Grid.js
sstrubberg: Yeppers. I couldn't figure out how to effect the column without introducing this.

packages/gatsby-theme-carbon/src/styles/internal/_page.scss
sstrubberg: Totally get it. I felt gross when i typed it ;)

packages/gatsby-theme-carbon/src/components/ResourceCard/ResourceCard.js
sstrubberg: Cancel that. Got it working with replace method.

packages/gatsby-theme-carbon/src/components/ResourceCard/ResourceCard.js
sstrubberg: figure it out. I'll show you tomorrow :)

PR #849, opened 2020-05-21
fix: ImageGalleryImage - change margin bottom of images from 1.5 rem …

xylish7:
The ImageGalleryImage margins were 32-X and 24-Y. This change makes the margins equal: 32-X 32-Y

- Old look

image

- New look

image

### Changelog
---
**Changed**

- margin-bottom of figure inside ImageGalleryImage to be 2rem instead of 1.5rem


vpicone:
Looks great, can we take this chance to use the carbon spacing spacing token?

I _think_ it’s `$spacing-08`

PR #857, opened 2020-05-29
chore: add table to components using <Title> tag

mariuszmickiewicz:
H4s and tables directly next to each other are too close visually. I think table should be mentioned here as another component which should use <Title> instead of H4 above it.

**Changed**

- added "table" to the copy


mariuszmickiewicz:
Example of a table with H4

image

Example of a table with <Title> 

image

vpicone:
@mariuszmickiewicz great point, this is the ideal use-case for the Title tag. Thanks for adding it!

PR #864, opened 2020-06-02
feat(Mini-resource-card)

### LINKED ISSUE #794: [Gatsby] Component Request: Add mini resource card to Theme

sstrubberg:
Closes #794

This component adds a `type` prop to the ResourceCard component. If the word "mini" is passed, then it will give you the new Mini ResourceCard, which has everything from the original ResourceCard gutted out except for the title and the children. The children are the icons that get passed. If no icons are passed as children, a default actionIcon of arrowRight will be given.

image

Thanks @jnm2377 for the idea to add a `gutterLg` prop to the `noGutterSm` when dealing with the mini sitting off to the side. This effectively kills the gutters for Sm and Md and then brings them back for Lg.

I also chose to ignore nested ternaries for the file. I do the nested ternaries in the `iconsAndImages` const. I read up on the reason why the linter flagged it and it's because of readability. I thought I could read it just fine, but I also know that I hold zero objectivity in this argument.

I feel pretty solid on the docs. I even added the calendar icon per a side convo with some friends on the accessibility team.


packages/gatsby-theme-carbon/src/styles/internal/_page.scss
```diff
@@ -134,6 +134,15 @@ img {
   padding-right: 0;
```
jnm2377, line 134: Or, on second thought (to my above comment https://github.com/carbon-design-system/gatsby-theme-carbon/pull/864/files#r434771615):

It this means specifically "there will be no gutter on small breakpoints", then I think this should be wrapped in a media query like so:

```css
.container .#{$prefix}--no-gutter-sm {
  @include carbon--breakpoint-down("sm") {
    padding-left: 0;
    padding-right: 0;
  }
}
```

Because then it means it's only being applied on sm breakpoints like the prop implies. Which means, that if the prop is excluded, it should have a gutter, right?

Then for aside minis, they would only need to add `noGutterSm` bc by default lg breakpoint will have a gutter...

jnm2377: I'm assuming that the default (w/o these noGutter styles) means that the column has a gutter.

packages/gatsby-theme-carbon/src/components/Grid.js
```diff
@@ -62,6 +63,7 @@ export const Column = ({
     [`${prefix}--no-gutter-lg`]: noGutterLg,
     [`${prefix}--no-gutter-xl`]: noGutterXl,
     [`${prefix}--no-gutter-max`]: noGutterMax,
+    [`${prefix}--no-gutter-sm--yes-gutter-lg`]: noGutterSmYesGutterLg,
```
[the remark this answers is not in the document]
sstrubberg: Ah! Great idea!

packages/gatsby-theme-carbon/src/styles/internal/_page.scss
jnm2377: @vpicone any thoughts on this?

vpicone: Typically the breakpoint stuff works is 'starting with this and all the way up' so you don't have to define it again and again through all the of the breakpoints.

The change above would add gutters to any breakpoints larger than small which might be an unexpected change.

jnm2377: Ah, I thought the prop `noGutterSm` meant no gutters only at that breakpoint. I see now, it means **starting** at that breakpoint.

packages/gatsby-theme-carbon/src/components/ResourceCard/ResourceCard.js
```diff
@@ -1,112 +1,184 @@
...
+    cardContainer = <div className={carbonTileclassNames}>{cardContent}</div>;
+  } else if (isLink === true) {
+    cardContainer = (
+      <Link to={href} className={carbonTileclassNames} {...rest}>
+        {cardContent}
+      </Link>
+    );
+  } else {
+    cardContainer = (
+      <a href={href} className={carbonTileclassNames} {...rest}>
```
jnm2377, line 143: if it's leaving the site, should we add a `target="_blank"` here? Otherwise, I'm not sure what the difference is between this and line 137.

packages/gatsby-theme-carbon/src/components/ResourceCard/ResourceCard.js
```diff
@@ -1,112 +1,184 @@
...
+            <h4 className={`${prefix}--resource-card__title`}>{title}</h4>
+          )
+        : null}
+      {type === 'mini'
+        ? title && (
+            <h4 className={`${prefix}--resource-card__mini-title`}>{title}</h4>
+          )
+        : null}
+    </>
+  );
```
jnm2377, line 88: Could we maybe simplify this? 
```suggestion
 const titleClassNames = classnames({
 [`${prefix}--resource-card__title`]: type === "default",
 [`${prefix}--resource-card__mini-title]: type === "mini",
 });
```

packages/gatsby-theme-carbon/src/components/ResourceCard/ResourceCard.js
```diff
@@ -1,112 +1,184 @@
...
+
+  // This holds the subtitle, title, icons and images.
+  const cardContent = (
+    <>
+      {subTitle && (
+        <h5 className={`${prefix}--resource-card__subtitle`}>{subTitle}</h5>
+      )}
+      {titleSize}
```
jnm2377, line 126: Related to my above comment
```suggestion
      {title && (<h4 className={titleClassNames}> {title} </h4> )}
```

packages/gatsby-theme-carbon/src/components/ResourceCard/ResourceCard.js
```diff
@@ -1,112 +1,184 @@
...
+      ) : null}
+      {type === 'mini' ? aspectRatioObjectType : null}
+    </>
+  );
+
+  return (
+    <div {...rest} className={ResourceCardClassNames}>
+      {aspectRatioClassNames}
+    </div>
+  );
```
jnm2377, line 180: Wondering if this too could be simplified. It can just be really confusing for maintainers when there are so many conditionals.

```suggestion
  // This determines, based whether the card is default or mini, whether or not it gets the aspect-ratio and aspect-ratio-X:X classes and associated div.
  const aspectRatioContainer = (
      {type === 'default' ? (
        <div
          className={`${prefix}--aspect-ratio ${prefix}--aspect-ratio--${aspectRatio.replace(
            ':',
            'x'
          )}`}
        >
          <div className={`${prefix}--aspect-ratio--object`}>
             {cardContainer}
           </div>
        </div>
      ) : cardContainer} // just gets cardContainer if type==="mini"
  );

  return (
    <div {...rest} className={ResourceCardClassNames}>
      {aspectRatioContainer}
    </div>
  );
```

jnm2377:
This is looking really great! ❤️ Just left some comments that I think could simplify some of the code for future maintenance.

andreancardona:
@sstrubberg this is looking so cool! I agree with @jnm2377 's suggestions :)

packages/gatsby-theme-carbon/src/components/ResourceCard/ResourceCard.js
sstrubberg: As I understand it, the else-if sets up the <Link> component so that when you navigate to a new page within the site, it does [preloading](https://www.gatsbyjs.org/docs/gatsby-link/) for a performance boost.. As opposed to the else which sets up the href, which will load the entire site.

packages/gatsby-theme-carbon/src/components/ResourceCard/ResourceCard.js
sstrubberg: I would love to get on a call with you guys and brainstorm on this one.

packages/gatsby-theme-carbon/src/components/ResourceCard/ResourceCard.js
vpicone: I think these would be more clear if you thought of them as components:

```jsx
const Title = () => {
	// return early if no title
	if !title return null;

	// set default class name
	let className = `${prefix}--resource-card__title`;

	// change className if we get a different type
	if (type === 'mini') {
		className = `${prefix}--resource-card__mini-title`;
	}

	return <h4 className={className}>{title}</h4>
}



// later
<Title />
```

If you wanna give the other ones a shot, we can collab on them monday as well

packages/gatsby-theme-carbon/src/components/ResourceCard/ResourceCard.js
vpicone: Yeah let's collab on this as a group, I think it's a great chance to practice together

packages/gatsby-theme-carbon/src/components/ResourceCard/resource-card.scss
```diff
@@ -40,7 +40,8 @@
   margin-bottom: 0;
 }
 
-.#{$prefix}--resource-card__icon--img .gatsby-resp-image-wrapper {
+.#{$prefix}--resource-card__icon--img .gatsby-resp-image-wrapper,
```
vpicone, line 43: Trying to avoid targeting the gatsby image wrapper. This has bitten us in the past when Gatsby changes this class or someone uses an image that is just an `img` tag wrather than markdown/gatsby-image

PR #870, opened 2020-06-03
feat(switcher): automate max height, closes on bg click, update disabled tab index

### LINKED ISSUE #840: [Gatsby theme] Top right Nav Switcher to also close when clicking on background.
Accessibility team recommend shaving the top right switcher also close when you click anywhere in the background and not just the close button in the top right.

### LINKED ISSUE #856: Programatically set the max-height for the switcher nav
Given users can replace Switcher content with their own, and we are semi-frequently [modifying Switcher content](https://github.com/carbon-design-system/gatsby-theme-carbon/pull/855) ourselves, we should opt for a more automated solution to animating the Switcher height.

One solution would be to use a DOM ref on the list and measure it's height, using this value to calculate the max-height of the nav wrapping it.

jnm2377:
Closes #840 
Closes #856 

As well as fixes a bug I found:
Screen Shot 2020-06-03 at 2 46 15 PM

**Changed**

- Closes switcher if you click elsewhere on the page (previously, this only happened on mobile, lg breakpoints only closed via the close button)
- Automates max-height styles
- Updates tabindex so that switcher items aren't focusable if they are disabled


packages/gatsby-theme-carbon/src/components/Switcher/Switcher.js
```diff
@@ -1,19 +1,38 @@
...
+  const [height, setHeight] = useState(null);
+
+  // calculate and update height
+  useEffect(() => {
+    switcherIsOpen
+      ? setHeight(listRef.current.offsetHeight + 40)
+      : setHeight(0);
+  }, [listRef, switcherIsOpen]);
+
+  const lgBreakpoint = window.matchMedia('(min-width: 1056px)'); // returns true if screen width matches media query
```
andreancardona, line 19: For some reason it looks like it doesn't like the `window`  in the build process?

Could be these things: 

- https://github.com/gatsbyjs/gatsby/issues/5835

- https://www.gatsbyjs.org/docs/debugging-html-builds/#how-to-check-if-code-classlanguage-textwindowcode-is-defined

BTW: let me know if you want to debug together

vpicone: @andreancardona yeah the build process is also known as 'pre-rendering' or 'server side rendering'. It tries to create as much of the html, css, and content as it can to be sent when the user first makes a request. 

When it's doing all this on your laptop, or GitHub there's no "window" it's just running as a command line script so it doesn't know what the value of window should be or if it meets that min-width.

You would want to use the little snippet they put there:

```js
if (typeof window !== `undefined`) {
  const module = require("module")
}
```

**But:** We have a hook for this! Do a project-wide search for `useMedia` or check out the `ArtDirection` component to see it in action

packages/gatsby-theme-carbon/src/components/Switcher/Switcher.js
```diff
@@ -1,19 +1,38 @@
...
+    }),
+  };
+
   return (
     <nav
       className={cx(nav, { [open]: switcherIsOpen })}
       aria-label="IBM Design ecosystem navigation"
       aria-expanded={switcherIsOpen}
       tabIndex="-1"
+      style={styles.container(lgBreakpoint.matches, switcherIsOpen)}
```
vpicone, line 33: I think we could clean this up by figuring out what the max-height is supposed to be outside of JSX above and then just passing it in like

```jsx
//outside of jsx
const maxHeight = calculationstuff


// later
style={{ maxHeight }}
```

packages/gatsby-theme-carbon/src/components/Switcher/Switcher.js
jnm2377: Nice! Didn't know we had a hook for it. Updated!

andreancardona: Ohh so cool! Good to know :)

packages/gatsby-theme-carbon/src/components/Switcher/Switcher.js
```diff
@@ -16,12 +17,10 @@ const Switcher = ({ children }) => {
...
   }, [listRef, switcherIsOpen]);
 
-  const lgBreakpoint = window.matchMedia('(min-width: 1056px)'); // returns true if screen width matches media query
+  const lgBreakpoint = useMedia('min-width: 1056px');
   const styles = {
-    container: (isLg, isOpen) => ({
-      /* eslint no-unused-expressions: [2, { allowShortCircuit: true, allowTernary: true }] */
-      maxHeight: !isLg && isOpen ? '100%' : `${height}px`,
-    }),
+    /* eslint no-unused-expressions: [2, { allowShortCircuit: true, allowTernary: true }] */
```
vpicone, line 22: What do you think about doing this in a way that gets around the eslint ignore? We could get rid of the `styles` object as well since it only has one property.

```js
let maxHeight = '100%';
if (!largeBreakpoint && switcherIsOpen) {
	maxHeight = `${height}px`;
}

//later
style={{maxHeight}}
```

jnm2377: @vpicone so doing it that way throws the same eslint error.

jnm2377: I tried this and still got the same error.

```js
  let maxHeight = `${height}px`;
  if (!lgBreakpoint && switcherIsOpen) {
    maxHeight = "100%";
  }
```

vpicone:
@jnm2377 the disabled thing is actually intended. You want disabled form elements to be tabbable. Otherwise users navigating by keyboard won't know they exist at all, which is a different thing than the "exists but isn't selectable" info that sited users get from a disabled element.

jnm2377:
@vpicone we don't do that in any of our carbon components. For example if a button is disabled, we it's not tabbable. How does that make sense here?

vpicone:
Like as a sites user I can see “Workplace” exists but is t ready yet. The idea is we want keyboard users to get that same info (i.e. workplace exists but it’s disabled)

packages/gatsby-theme-carbon/src/components/Switcher/Switcher.js
vpicone: Sorry it took me so long to get to this, lets take a look at it together whenever you have a sec.

jnm2377:
> @jnm2377 I think you're right just because they're disabled doesn't mean they're invisible to screen readers. Not sure if there's any screen reader hints that the link is disabled or if it'd show up as normal text. But either way I think taking them out of the tab order is the right move here.

@dakahn just want to confirm this is ok to do. We currently have a disabled item in the switcher and it's still tabbable. In this PR, I updated switcher items to not be tabbable if they're disabled. Is that ok? It seems like that's the pattern we do for everything else that's disabled.

vpicone:
@jnm2377 yup, I'm on board with your change.

It's weird because there is no 'disabled' attribute for links (like there is for buttons), so its sort of a weird gray area but I think sticking to the Carbon behavior (not focusable when disabled) is the right move for sure.

jnm2377:
@vpicone looks like the changes are working correctly 🎉

packages/gatsby-theme-carbon/src/components/Switcher/Switcher.module.scss
```diff
@@ -36,7 +35,6 @@
     // https://css-tricks.com/using-css-transitions-auto-dimensions/
     // Tied to the height of the app switcher. Allows for animating up to auto.
     // Matching the height exactly is imperitive to keep the correct animation timing.
```
vpicone, line 37:
```suggestion
```

vpicone:
🚀

PR #881, opened 2020-06-17
feat(MiniCard)

### LINKED ISSUE #794: [Gatsby] Component Request: Add mini resource card to Theme

sstrubberg:
Closes #794 

- Two new components `<CardGroup>` and `<MiniCard>`. Navigate to the `<MiniCard>` page to see how they are implemented.
- Touched up the `<ResourceCard>` as well, added a little diddy for replacing the ":" with an "x".
- Removed a non-existent className from the `<ArticleCard>`

Thanks @jnm2377 for the idea to add a gutterLg prop to the noGutterSm when dealing with the mini sitting off to the side. This effectively kills the gutters for Sm and Md and then brings them back for Lg.

I feel pretty solid on the docs. I even added the calendar icon per a side convo with some friends on the accessibility team.


packages/gatsby-theme-carbon/src/components/CardGroup/CardGroup.module.scss
```diff
@@ -0,0 +1,49 @@
...
+
+.row
+  [class*="--col-lg"]:nth-child(-n + 2)
+  .#{$prefix}--resource-card.#{$prefix}--resource-card--dark {
+  border-top-color: $carbon--black-100;
+}
+
+// Borders for stacking MiniCards within a column
+
+.row [class*="--col-lg"] a:nth-child(n + 2) [class*="MiniCard-module--card--"] {
```
vpicone, line 35: Ideally we don't wanna use the `[class*=` technique for rules we define in the theme. That's a good escape hatch for theme sites but in the theme itself, we want to avoid depending on class names generated in different modules.

One option would be having replacing this with a generic `.groupable-card` class and importing it into the various cards that can be used with the card group. This also allows someone to use `CardGroup` with a card they've defined rather than the one's we've hard-coded rules for.

In that same vein, we generally don't want to use elements in our styles if we can avoid it. There's no guarantee a card will be an anchor tag. We can use a wildcard there to say we're down with buttons, links, images, whatever.

```suggestion
.row [class*="--col-lg"] *:nth-child(n + 2) .groupable-card {
```

packages/gatsby-theme-carbon/src/components/MiniCard/MiniCard.js
```diff
@@ -0,0 +1,71 @@
...
+    <>
+      {actionIcon === 'launch' ? <Launch20 aria-label="Open resource" /> : null}
+      {actionIcon === 'arrowRight' ? (
+        <ArrowRight20 aria-label="Open resource" />
+      ) : null}
+      {actionIcon === 'download' ? <Download20 aria-label="Download" /> : null}
+      {actionIcon === 'email' ? <Email20 aria-label="Email" /> : null}
+      {actionIcon === 'calendar' ? <Calendar20 aria-label="Calendar" /> : null}
+    </>
+  );
```
vpicone, line 34: I think the clearest way to do this would be with a `getIcon` helper function outside of the MiniCard that uses a switch statement to return the correct icon:

```suggestion
const getIcon = ({actionIcon}) => {
  switch (actionIcon) {
    case 'arrowRight':
      return <ArrowRight20 aria-label="Open resource" />
    case 'download':
        return <Download20 aria-label="Download" />;
    case 'email':
        return <Email20 aria-label="Email" />;
    case 'calendar':
        return <Calendar20 aria-label="Calendar" />;
    default:
        return <Launch20 aria-label="Open resource" />;
  }
}
```

packages/gatsby-theme-carbon/src/components/MiniCard/MiniCard.js
```diff
@@ -0,0 +1,71 @@
...
+
+  const cardContent = (
+    <div className={styles.card}>
+      <div className={styles.wrapper}>
+        <div className={styles.title}>{title}</div>
+        {children === undefined ? (
+          <div className={styles.icon}>{icons}</div>
```
vpicone, line 41: and then call the helper later

```suggestion
          <div className={styles.icon}>{getIcon({actionIcon})}</div>
```

packages/example/src/pages/components/MiniCard.mdx
```diff
@@ -0,0 +1,114 @@
+---
+title: MiniCard
+description: Usage instructions for the MiniCard component
+---
+
```
vpicone, line 5: @jeanservaas can you provide some context on when someone would reach for this card as opposed to a resource card? It'd be great to have some design intent rather than just having folks choose. Some sites/projects might not have a designer to pick between the two.

packages/gatsby-theme-carbon/src/components/CardGroup/CardGroup.module.scss
vpicone: consider flattening rules, ideal: `.row .groupable-card:nth-child(n+2)`

vpicone:
@jeanservaas @mjabbink we're pretty close here but need some guidance on when a designer would reach for this instead of a regular resource card.

packages/example/src/pages/components/MiniCard.mdx
sstrubberg: > Dope! Coming together. If I remember correctly, we nuked the `Children.map` stuff and settled on moving the column into the MiniCard component (ala DoDont).
> 
> That way each card could get their own column props rather than forcing them all to use the same (like it would be if we applied them with Children.Map)

Yeah, decided to keep the Children.map and keep the Column with the CardGroup because when i remove the Children.map and move the Column to the MiniCard itself it started to break down the aside case that Jeannie was looking for.

sstrubberg:
> Dope! Coming together. If I remember correctly, we nuked the `Children.map` stuff and settled on moving the column into the MiniCard component (ala DoDont).
> 
> That way each card could get their own column props rather than forcing them all to use the same (like it would be if we applied them with Children.Map)

Yeah, decided to keep the Children.map and keep the Column with the CardGroup because when i remove the Children.map and move the Column to the MiniCard itself it started to break down the aside case that Jeannie was looking for.

vpicone:
@sstrubberg oooo interesting, does it not work if you just pass these column props through? https://github.com/carbon-design-system/gatsby-theme-carbon/pull/881/files#diff-a6e4adfabe57f38e4aef76cc5eb11485R47

sstrubberg:
> @sstrubberg oooo interesting, does it not work if you just pass these column props through? https://github.com/carbon-design-system/gatsby-theme-carbon/pull/881/files#diff-a6e4adfabe57f38e4aef76cc5eb11485R47

Yeah, it doesn't. When I move the column over to the MiniCard itself, it brings the column (group) styling with it. The aside minicard starts to be treated like a cardgroup.

jeanservaas:
@vpicone you've raised some really good points about this mini card and we've already had a lot of confusion about its use. 

The fundamental problem with the mini resource card is that, unlike the larger resource card, it can only have one icon. Where the large resource cards have two icons a 32px icon and a 20px arrow. 

image


It was originally inspired by these mini resources cards that cross link on material.io to related articles

image

When we remove the 32px icon, we're left, basically, with something that resembles a button almost exactly. 

@mjabbink and I talked this and I think, the only way to (potentially) salvage this guy is to:

- [ ]  remove the arrows as an option on the mini resource card (i.e. reference the larger icon, from the big resource card, not the smaller arrow).

- [ ] bump up the size of the icon to 24px (I'm on the fence about going to 32px for consistency, but it will probably be too horsey)

image

If we use it to cross link to a related component / pattern on the site, we can use the Carbon icon instead of the arrow.

**(Looking at this PR, I think maybe @sstrubberg I need to make a separate issue for this mini card enhancement. I just don't think we should merge mini card in without these updates). Then I will write a few lines about not using it for actions and maybe add a couple of usage images for it.**

mjabbink:
I guess the key word is resource. Action buttons are typically not aiming big to knowledge/education/assets/guidance. But yes it’s a thin difference

packages/gatsby-theme-carbon/src/components/MiniCard/MiniCard.module.scss
```diff
@@ -0,0 +1,32 @@
...
+  @include carbon--type-style("body-long-02");
+}
+
+.image {
+  width: 20px;
+  height: 20px;
+}
+
+.icon {
+  padding-top: 4px;
```
vpicone, line 32: I think this is `$spacing-02` not sure if we have `20px` for the image, we should totally document the spacing tokens better haha

vpicone:
@sstrubberg lmk if you need help with the merge dealio

packages/gatsby-theme-carbon/src/components/MiniCard/CardGroup.js
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+import { Row, Column } from '../Grid';
+import { cardGroup } from './CardGroup.module.scss';
+
+const CardGroup = ({ children, className, ...rest }) => (
```
vpicone, line 5: Hmm I thought we would just give them a cardGroup class? Can't we just do 
```css
.cardGroup .column
```
So we don't have to map over the children? Following the pattern of DoDontRow

packages/example/src/pages/components/MiniCard.mdx
```diff
@@ -0,0 +1,118 @@
+---
+title: MiniCard
+description: Although the mini-resource card has a similar geometry to the button component, they should not be used in place of a button. Buttons encourage action from the user and affect the website's front-end or back-end. The resource cards, both large and mini are essentially links. They are used for navigation and actions that do not affect the website.
```
jnm2377, line 3: this explanation is 🔥

mjabbink:
@jeanservaas I think the type inside mini is supposed to be 14px. All the current use cases are 14px.

packages/example/src/pages/components/MiniCard.mdx
sstrubberg: that's all @jeanservaas ;)

sstrubberg:
> @jeanservaas I think the type inside mini is supposed to be 14px. All the current use cases are 14px.

it's currently using `body-long-02`

packages/gatsby-theme-carbon/src/styles/internal/_page.scss
```diff
@@ -150,6 +150,13 @@ img {
   padding-left: 0;
 }
 
+.container .#{$prefix}--gutter-lg {
```
vpicone, line 153: Should these go in the column module instead?

jeanservaas:
@sstrubberg Mike's right, type in the mini should match the larger resource card type and be $body-short-01 (14px / 18px). My bad in the spec.

PR #893, opened 2020-06-29
fix: ordered lists use native list style to allow for arbitrary start number

### LINKED ISSUE #495: Nested list item width
Continuation of #494

- [ ] figure out what things, if any, should be list-width and what should be full width (minus the list's indentation).
- [ ] better document how to nest MDX elements within lists

Screen Shot 2019-08-12 at 4 39 10 PM
Screen Shot 2019-08-12 at 4 38 50 PM

Screen Shot 2019-08-12 at 4 37 01 PM
Screen Shot 2019-08-12 at 4 36 19 PM

jnm2377:
Closes #495 

Changes ordered lists style to use native `decimal` style instead of `::before` element, so that we can have lists start with an arbitrary number using the native `start` attribute for `ol`.

View example at `/test/spacing-audit/`


peter-garvin:
Its beautiful

peter-garvin:
its everything ive ever dreamed of

packages/gatsby-theme-carbon/src/components/markdown/Markdown.module.scss
```diff
@@ -169,6 +174,14 @@
...
   color: $carbon--blue-40;
 }
+
+:global(.bx--list--ordered:not(.bx--list--nested) > .bx--list__item::before) {
+  display: none;
+}
+
+:global(.bx--list--ordered:not(.bx--list--nested)) .list-item {
+  margin-left: 0.5rem;
+}
```
vpicone, line 184: Could we combine these two selectors, or target them in a way that doesn't require global styles?

PR #895, opened 2020-07-01
fix(ResourceCard): download attribute, fixes #871

### LINKED ISSUE #871: Downloading files from '/files' directory gives initial 404 screen
## Detailed description

I’m having an unexpected issue with the ResourceCard component -

Scenario:
In develop mode -

I have PDFs, Presentation and other document files in the `/static/files` directory. These files should be available from `https://host:port/files`.

I’m using the resource card component -
```mdx
<ResourceCard
subTitle="Powerpoint presentation"
href="/files/Operator.pptx"
>

![Github icon](../images/github-icon.svg)

</ResourceCard>
```

When I click on the resource card for the presentation on this [GitHub Enterprise page - GHE](https://pages.github.ibm.com/Dawn-N-Ahukanna/ai-apps-persona-prototype/ai-apps/files/Bryan-Operator.pptx), I get a 404 error and when I refresh the 404 page, the file download shows up. Why am I initially getting the 404 page and only on page fresh getting the file download dialog?

Screenshot 2020-05-29 at 20 01 15

I'd expect to get the file download dialog or for the file to download to the default download location for my browser.

* Firefox for developers 78.0b2 (64-bit)
* Firefox 77.0.1 (64-bit)
* Chrome 83.0.4103.61 (Official Build) (64-bit)

1. Go to [GHE page](https://pages.github.ibm.com/Dawn-N-Ahukanna/ai-apps-persona-prototype/ai-apps/personas/maximo/bryan#other-resources)
2. Click on resource component - [Link](https://pages.github.ibm.com/Dawn-N-Ahukanna/ai-apps-persona-prototype/files/Bryan-Operator.pptx)
3. Get 404 page with '/files/Bryan-Operator.pptx'
4. Refreshing the page shows the file download dialog - 
Screenshot 2020-05-29 at 20 01 15

## Additional information
None.

se4m0re:
Check if download attribute is present.
Call withPrefix to prepend path prefix to href if download attribute
present.

Closes #871 

The ResourceCard component checks the first char given in the url (`isLink = href.charAt(0) === '/'`), treating it as an internal link if the first char is a `/` and passes it to the `<Link>` component. This created an issue in one particular use case where the ResourceCard would be pointing to a downloadable asset found inside the `/static` folder at the root of the project. Checking to see if a download prop is present solves this issue.

### DIFF
--- a/packages/gatsby-theme-carbon/src/components/ResourceCard/ResourceCard.js
+++ b/packages/gatsby-theme-carbon/src/components/ResourceCard/ResourceCard.js
@@ -89,13 +89,15 @@ export default class ResourceCard extends React.Component {
         </Link>
       );
     } else {
+      // The URL is assumed to be external if an http protocol is present or defaulted (i.e. it is a protocal-relative URL).
+      const isExternalURL = RegExp(/^(https?:)?\/\//g).test(href);
+
+      // Prepend the path prefix in production.
+      const hrefPrefixed = withPrefix(href);
+
       cardContainer = (
         <a
-          href={
-            rest.download && !RegExp(/^(https?:)?\/\//g).test(href)
-              ? withPrefix(href)
-              : href
-          }
+          href={rest.download && !isExternalURL ? hrefPrefixed : href}
           className={carbonTileclassNames}
           {...rest}
         >
--- a/packages/gatsby-theme-carbon/src/components/ResourceCard/ResourceCard.js
+++ b/packages/gatsby-theme-carbon/src/components/ResourceCard/ResourceCard.js
@@ -1,7 +1,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import classnames from 'classnames';
-import { Link } from 'gatsby';
+import { Link, withPrefix } from 'gatsby';
 import {
   Launch20,
   Download20,
@@ -29,7 +29,7 @@ export default class ResourceCard extends React.Component {
     } = this.props;
 
     let isLink;
-    if (href !== undefined) {
+    if (href !== undefined && !rest.download) {
       isLink = href.charAt(0) === '/';
     }
 
@@ -90,7 +90,11 @@ export default class ResourceCard extends React.Component {
       );
     } else {
       cardContainer = (
-        <a href={href} className={carbonTileclassNames} {...rest}>
+        <a
+          href={rest.download ? withPrefix(href) : href}
+          className={carbonTileclassNames}
+          {...rest}
+        >
           {cardContent}
         </a>
       );
--- a/packages/gatsby-theme-carbon/src/components/ResourceCard/ResourceCard.js
+++ b/packages/gatsby-theme-carbon/src/components/ResourceCard/ResourceCard.js
@@ -91,7 +91,11 @@ export default class ResourceCard extends React.Component {
     } else {
       cardContainer = (
         <a
-          href={rest.download ? withPrefix(href) : href}
+          href={
+            rest.download && !RegExp(/^(https?:)?\/\//g).test(href)
+              ? withPrefix(href)
+              : href
+          }
           className={carbonTileclassNames}
           {...rest}
         >
--- a/.gitignore
+++ b/.gitignore
@@ -71,3 +71,4 @@ yarn-error.log
 # Vscode settings
 .vscode/settings.json
 .vscode/tasks.json
+.vercel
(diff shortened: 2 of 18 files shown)


packages/gatsby-theme-carbon/src/components/ResourceCard/ResourceCard.js
```diff
@@ -90,7 +90,11 @@ export default class ResourceCard extends React.Component {
       );
     } else {
       cardContainer = (
-        <a href={href} className={carbonTileclassNames} {...rest}>
+        <a
+          href={rest.download ? withPrefix(href) : href}
```
vpicone, line 94: Hey! Thanks for the PR. This won't work if the download is from a different site, it will apply the path prefix for the current page.

se4m0re: Good point Vince. Maybe testing to see if an http protocol is present (or if it is being defaulted) at the beginning of the string would be a sufficient check on whether the resource is internal or external.

```
var href = 'https://1.dam.s81c.com/m/7e6c4e33b202794c/original/06222020-ls-watson-works-2304x983.jpg'
RegExp(/^(https?:)?\/\//g).test(href)  //true

href = 'http://1.dam.s81c.com/m/7e6c4e33b202794c/original/06222020-ls-watson-works-2304x983.jpg'
RegExp(/^(https?:)?\/\//g).test(href)  //true

href = '//1.dam.s81c.com/m/7e6c4e33b202794c/original/06222020-ls-watson-works-2304x983.jpg'
RegExp(/^(https?:)?\/\//g).test(href)  //true

href = '/1.dam.s81c.com/m/7e6c4e33b202794c/original/06222020-ls-watson-works-2304x983.jpg'
RegExp(/^(https?:)?\/\//g).test(href)  //false

href = '/http://1.dam.s81c.com/m/7e6c4e33b202794c/original/06222020-ls-watson-http://works-2304x983.jpg'
RegExp(/^(https?:)?\/\//g).test(href)  //false


```

se4m0re: The code below seems to handle the example I posted previously. However, Chrome and Firefox (and probably most browsers) only support same origin download links (https://caniuse.com/#feat=download). Testing the above url for the image showed that Chrome loads the image in current browser tab, which could be changed to open in a new tab by supplying `target="_blank"`.

```
    let cardContainer;
    if (disabled === true || href === undefined) {
      cardContainer = <div className={carbonTileclassNames}>{cardContent}</div>;
    } else if (isLink === true) {
      cardContainer = (
        <Link to={href} className={carbonTileclassNames} {...rest}>
          {cardContent}
        </Link>
      );
    } else {
      cardContainer = (
        <a
          href={
            rest.download && !RegExp(/^(https?:)?\/\//g).test(href)
              ? withPrefix(href)
              : href
          }
          className={carbonTileclassNames}
          {...rest}
        >
          {cardContent}
        </a>
      );
    }
```

vpicone:
Could we instead just provide the full URL for downloadable files?

se4m0re:
Why not utilize the `withPrefix` helper function inside the component instead of having to supply the full URL? If the path prefix needed to be changed, you would not have to remember (and then make the effort) to go and change all the places where it's been hard coded. As long as the resource to be downloaded can be identified as either internal or external, adding the `withPrefix` helper shouldn't be problematic.

packages/gatsby-theme-carbon/src/components/ResourceCard/ResourceCard.js
```diff
@@ -90,7 +90,15 @@ export default class ResourceCard extends React.Component {
       );
     } else {
       cardContainer = (
```
vpicone, line 92:
```suggestion
	  // move the href logic up here and comment what it's doing
      cardContainer = (
```

sstrubberg:
Does the ResourceCard documentation need to be updated to reflect these changes involving the download attribute?

vpicone:
@sstrubberg I think it should 'just work'

packages/gatsby-theme-carbon/src/components/ResourceCard/ResourceCard.js
se4m0re: Okay, I think it's a bit easier to read/understand now.

vpicone: Agreed! Thank you.

PR #900, opened 2020-07-07
Chores

vpicone:
This does a collection of minor chores
1. fix the class/className warning
2. update deps
3. fix a new eslint a11y warning that cropped up


packages/gatsby-theme-carbon/gatsby-ssr.js
```diff
@@ -9,7 +9,7 @@ export const wrapRootElement = wrapRoot;
 export const onRenderBody = ({ setHeadComponents, setBodyAttributes }) => {
   // prevents jank while page hydrates, removed on page load
   setBodyAttributes({
-    class: 'body--preload',
+    className: 'body--preload',
```
sstrubberg, line 12: Gah! There it was! For the life of me.. I could not find where that error was coming from.

PR #901, opened 2020-07-07
fix: check for pathPrefix in video component to avoid path duplication

### LINKED ISSUE #896: Video component throws 404 when using MDX, github pages, and pathPrefix
I noticed on an IBM PAL site I was working on that the Video component wasn't properly working when deployed to github pages. It has since been fixed, but here's the page I was having difficulties with https://pages.github.ibm.com/cdai-design/guidance/walkme-guide/smart-walk-thrus/steps.

The documentation for using Video in MDX is laid out as such

```
import localVideo from './local-video.mp4';
import localPoster from './local-poster.png';
<Video src={localVideo} poster={localPoster} />
```

This does work for local development, but when pushed to github pages it was causing a 404 when i tried to implement.

```
import localVideo from './videos/video.mp4';
...
<Video src={localVideo} />
```

I compared the video URL to another static asset and this is what i was seeing

```
https://pages.github.ibm.com/cdai-design/guidance/cdai-design/guidance/static/SWT-steps-How-to_pagination-e0b43a3a4a19bda72219814238d67a17.mp4

https://pages.github.ibm.com/cdai-design/guidance/static/81ed67a9d6296a40dd263da1ddd38e04/4ea69/WalkMe-SWT-steps1.png
```

Part of the production URL was being duplicated `/cdai-design/guidance/cdai-design/guidance/`. I was able to deduce that this was being caused by this line from the `gatsby-config.js` file in the PAL repo `pathPrefix: '/' + (process.env.ENV === 'PROD' ? cdai.repo : ''),`. the value of `cdai.repo` being `cdai-design/guidance`. I didn't set up any of this, so i'm not 100% sure why it's there or if it's necessary. I haven't been able to ask the repo maintainer why it's there. Regardless, once I switched that if statement around it started working. I believe it's because that prefix is used in the video component on this line https://github.com/carbon-design-system/gatsby-theme-carbon/blob/master/packages/gatsby-theme-carbon/src/components/Video/Video.js#L107

The solution for me was to simply change my implementation to this `<Video src="/video.mp4" />` and move the file to a static folder at the top level of the repo. the source then linked correctly to that static folder. I'm not sure if this is strictly a bug or a weird edge case or what, but my only suggestion is that the documentation around the video component should be updated with a troubleshooting section that mentions this use case and how to fix it.

davidmenendez:
Closes #896

the src prop in the video component should be checked to see if it already contains the pathPrefix to avoid bug where pathPrefix was being applied twice and causing 404s

#### Changelog

**New**

- none

**Changed**

- Video.js

**Removed**

- none


davidmenendez:
@sstrubberg I gave as much information as I could in the issue. based on what i read in https://www.gatsbyjs.org/docs/path-prefix/ about `pathPrefix` usage it seems like usage of `withPrefix` in the video components `src` prop (https://github.com/carbon-design-system/gatsby-theme-carbon/blob/master/packages/gatsby-theme-carbon/src/components/Video/Video.js#L107) is causing the issue of needlessly appending the `pathPrefix` when importing like `import localVideo from './local-video.mp4'` thus causing a broken link. since `pathPrefix` is mentioned here https://gatsby-theme-carbon.now.sh/guides/npm-scripts/#build I would think that there is a need by users to support this. i'll also note that `withPrefix` is only used in the video component. perhaps that's not the correct solution in this instance?

my intention was just to raise awareness of an issue that I ran into while trying to develop with carbon theme for https://pages.github.ibm.com/cdai-design/guidance/

vpicone:
I think this is an issue with the Video component usage of withPrefix. We can just check to make sure the prefix isn’t already applied (as it is when someone imports a local video) before we use the withPrefix helper.

vpicone:
Using the static directory should be a last resort as it’s usually not cached properly.

davidmenendez:
@vpicone just to be clear I should revert the documentation changes and explore updating the video component directly?

vpicone:
@davidmenendez If you don't mind, I think that would be the best solution here. Sorry for going back and forth. I have a good idea of how it'd work and how to test it, let me know if you need any help or have any questions!

We have a `usePathPrefix` hook you can import from `src/util/hooks/usePathPrefix`. From there we can:
1. Check if they're using a prefix
2. Check if the src is missing the prefix
3. if `1.` and `2.` are true, use the `withPrefix` helper from Gatsby

To test this in the theme repo, you'll want to make sure it works with `yarn develop`, `yarn serve`, and especially `yarn test:prefix`

.The last two will build and serve the site locally with and without a path prefix respectively.

davidmenendez:
@vpicone updated code based on your feedback 👍

vpicone:
🚀 thanks for working with us on this, I feel like we landed on the right solution.

PR #902, opened 2020-07-08
Row padding prop

vpicone:
This adds a prop to col and row so we can control padding when wrapped

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -1,6 +1,8 @@
 {
-  "packages": ["packages/*"],
-  "version": "1.25.0",
+  "packages": [
+    "packages/*"
+  ],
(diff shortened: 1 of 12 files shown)


sstrubberg:
You might have to walk me through this. I brought your PR down to my workstation and i'm failing to see how this is helping groupings of ResourceCards and ImageCards.

vpicone:
@sstrubberg hmm yeah I'll leave this up for now for reference, but let's brainstorm tomorrow or something.

I think Ideally we could make the `CardGroup` work for `ImageCard` and `ResourceCard` without removing the `class="resource-card-group"` functionality. Then, just update the docs to use the new, preferred strategy.

Issue #906, opened 2020-07-09
Normalize Column usage for Card components


vpicone:
By making developers provide the Rows and Columns, we're unable to style or otherwise influence the behavior of rows/columns containing specific cards. We can solve this with custom wrappers for those Cards (ala DoDontRow), but we should look for a solution that allows a more generic card container (perhaps `CardGroup`).

We should aim to converge on a single API for components that allows them to take column props when appropriate. This prevents the boilerplate necessary to render card groups (pasting rows and columns with specific props everywhere). In addition it could eliminate the `resource-card-group` and `image-card-group` Rows that are a big gotcha when authoring groups of components.

PR #907, opened 2020-07-09
Col wrap fixes

vpicone:
## Changes
Made tweaks to individual component styles in order to account for the new default column spacing on mobile introduced by #899

Also created #906 as the first `v2.0` milestone issue. There we can make some breaking changes to normalize the column API across components and handle these sort of component cluster/grouping issues more holistically without tweaking individual components.

## Testing
Run through the components on desktop and mobile width, looking for spacing irregularities

## Release notes

We've updated the default column behavior to add vertical spacing when a row wraps (on mobile). If you've created a custom component using columns and would prefer that space not exist, just tweak the `--space` CSS custom property for the column like so:

```css
.my-special-column {
	--space: 0;
}
```

#### ImageCard
Be sure the `Row` containing your `ImageCard` components has the `image-card-group` class [from the code example.](https://gatsby-theme-carbon.now.sh/components/ImageCard#code) You might already have it. This class wasn't used before this release, but is used now to override the new column space and keep things tight. In the future, we'll be normalizing this sort of behavior in a generic `CardGroup` component (#906).

#### ResourceCard (shadowing)
`resource-card-group.scss` and `resource-card.scss` have been combined into `ImageGallery.scss`, this only matters if you're shadowing them; you'll need to combine them as well.

#### ImageGallery (shadowing)
`ImageGalleryImage.module.scss` and `ImageGallery.module.scss` have been combined into `ImageGallery.scss`, this only matters if you're shadowing them; you'll need to combine them as well.


packages/gatsby-theme-carbon/src/components/ResourceCard/resource-card-group.scss
```diff
@@ -1,4 +1,11 @@
 // Card groups
+
+// #906: We need to use the attribute selector here b.c. we don't control the column,
+// it's provided by the developers through MDX and we don't want to force an extra prop
+.resource-card-group [class*="--column"] {
```
sstrubberg, line 5: You can avoid the wildcard with this:
```
.resource-card-group .#{$prefix}--col-lg-4 {
  margin-bottom: 0;
}
```

packages/gatsby-theme-carbon/src/components/ImageCard/image-card.scss
```diff
@@ -1,3 +1,9 @@
+// #906: We need to use the attribute selector here b.c. we don't control the column,
+// it's provided by the developers through MDX and we don't want to force an extra prop
+.image-card-group [class*="--column"] {
```
sstrubberg, line 3: Same as my comment on ResourceCard.

```
.image-card-group .#{$prefix}--col-lg-4 {
  margin-bottom: 0;
}
```

packages/gatsby-theme-carbon/src/components/ResourceCard/resource-card-group.scss
vpicone: I think the `--column` approach is marginally better since we provide that class (as opposed to Carbon), and if we ever let in column props, we can't guarantee it'll be 4 columns but we can guarantee it'll have the `--column` class on it.

PR #909, opened 2020-07-11
Add carbon prettier

vpicone:


sstrubberg:
`prettier-config-carbon`. Super cool. Should we be including this in all our projects?

vpicone:
@sstrubberg I think so, super easy to implement and gets everyone on the same page.

PR #914, opened 2020-07-16
chore: change default spacing

vpicone:
Updates the default spacing from 24px to 32px


mjabbink:
@vpicone it’s the same as the margin on the right. @sadekbazaraa was thinking these need to be 64px. @sadekbazaraa please chime in on this because we either default to 32px and add 32px because we have situations where resource cards stack under that need to be 32px. This seems to be the right strategy.

PR #948, opened 2020-07-28
chore: manually apply muted attribute

vpicone:
React eats the muted attribute (https://github.com/facebook/react/issues/10389#issuecomment-605689475), this is a safety measure to ensure that this attribute is applied and autoplay works as expected on mobile devices.


sstrubberg:
Do we have a video with audio on the site that we can test this against?

vpicone:
@sstrubberg should be able to use the non-vimeo videos here https://gatsby-theme-carbon.now.sh/components/Video

Issue #966, opened 2020-08-13
Accessibility Issue: GlobalSearch does not allow ESC press to close the search/siteSwitcher


aliunwala:
This issue is also being tracked in our project here https://github.ibm.com/IBMa/E2E/issues/963
We see this issue on the https://ibm.com/able site as well as we are importing the Gatsby GlobalSearch implementation.

Title: GlobalSearch does not allow ESC press to close the search/siteSwitcher

### Describe in detail the issue you’re having.
Basically pressing the ESC key does not close the drop down menus (search/siteSwitcher) on the GlobalSearch component

Here is the location:
![image](https://user-images.githubusercontent.com/4451128/90093461-4751e580-dcf1-11ea-84dc-ecb301577327.png)

### What did you expect to happen? What happened instead? What would you like to see changed?
The menu to close on ESC key press

### What browser are you working in?
Browser agnostic

## Steps to reproduce the issue
Tab several times and press ESC(I am pressing ESC over an over and the drop downs are staying open):
![EscBug](https://user-images.githubusercontent.com/4451128/90093633-a7e12280-dcf1-11ea-81b2-e09b13d5ca34.gif)

## Fix
I have a fix for this one that I put into the PR here: https://github.com/carbon-design-system/gatsby-theme-carbon/pull/967

I am tabbing around and pressing ESC to close the dropdowns:
![EscBugFixed](https://user-images.githubusercontent.com/4451128/90094279-1ecaeb00-dcf3-11ea-9128-99e57d2330c2.gif)

I trigger on the esc keyup to set the state of the switcherIsOpen/searchIsOpen to be closed. This seems to work well when I tested it locally after building the code. 

There is one improvement that might be worth doing. With the proposed fix the ESC key has to be pressed twice to get out of the search. So if you know @vpicone  how this might be fixed/made better let me know and ill update the PR accordingly.

PR #967, opened 2020-08-13
Accessibility bug press esc to close global search

aliunwala:
See ticket here:
https://github.com/carbon-design-system/gatsby-theme-carbon/issues/966


packages/gatsby-theme-carbon/src/components/GlobalSearch/GlobalSearch.js
```diff
@@ -180,6 +180,12 @@ const GlobalSearchInput = () => {
             onClick={() => {
               toggleNavState('searchIsOpen', 'open');
               toggleNavState('switcherIsOpen', 'close');
+              document.addEventListener('keyup', function(e){
```
vpicone, line 183: We'll want to move this to a `useEffect` hook. Putting it here means the event listener will be added to the document every time the button is clicked.

vpicone: @aliunwala do you need some assistance with this? it's a little reactspecific

aliunwala: I'll give it a try shortly and see if I can get it to work. Thanks for reaching out.

packages/gatsby-theme-carbon/src/components/GlobalSearch/GlobalSearch.js
```diff
@@ -78,6 +78,15 @@ const GlobalSearchInput = () => {
     [clearAndClose, focusedItem]
   );
 
+  useEffect(() => {
```
aliunwala, line 81: @vpicone Here is my attempt to try to wrap that code in the useEffect hook. BEWARE: this code freezes the page. So do not merge it. Is this the right way to use the useEffect hook? I am relatively new to this react hook.

aliunwala: Some more detail. The website freezes after the ESC key is pressed a few times on any page. The site renders fine and you are able to move around the site normally.

vpicone:
@aliunwala sorry for not getting to this sooner! Checking it out now.

aliunwala:
Not a problem @vpicone,  just let me know how I can help, I am happy to update the current fix (With the useEffect hook)

packages/gatsby-theme-carbon/src/components/GlobalSearch/GlobalSearch.js
vpicone: You're missing the dependencies and cleanup part of the `useEffect` hook. Check out the docs here: https://reactjs.org/docs/hooks-reference.html#useeffect

sstrubberg:
@aliunwala You still working on this or shall we close it for now?

aliunwala:
@sstrubberg Hey Scott, this fell off my radar, thanks for reminding me. I'll take a look at this this afternoon and make an update to the PR.

packages/gatsby-theme-carbon/src/components/GlobalSearch/GlobalSearch.js
aliunwala: That was very helpful, by adding the cleanup part I was able to get the site to work "as expected/normally" with the additional esc functionality. 

@vpicone take a look at this one and see if it is ready to get merged.

aliunwala: Doing a bit more testing... I think their still is an issue:
image

This seems to happen when there is text in the search bar when esc is pressed

When there is text in the field and the esc key is pressed what would be the expected behavior? Delete the text? Or save the state?

vpicone: @aliunwala I don't think escape should have any impact on the search. If you reduce your browser width to mobile sizes, you'll see [MDN](https://developer.mozilla.org/en-US/) doesn't perform any actions there.

My hunch is it would clear the search, then if the search is clear, it would close it.

vpicone:
@aliunwala I moved the listener to the Switcher, which should definitely close when escape is pressed. However, given the MDN example and some other research, I've not convinced clearing and closing the search based on the escape key is a settled pattern. 

We already clear and close when search is active (the user's cursor is in the input), changing the state and content of inputs when the user isn't focused on the input can result in us changing the value without the user knowing. I think that might actually be worse for accessibility considering the user can just tab out or tab to the close button as-is.

PR #975, opened 2020-08-19
fix: positioning of ImageCard on mobile view

xylish7:
#### Changelog

**Changed**

- the position of ImageCard on mobile view by removing the white space on the right side

Before fix:

image

After fix:

image


xylish7:
From what I saw that `-1px` is used to create that gap both on bottom and right. Taking in consideration that the span container has `position: relative` it will move the image 1px.  What I've done was to not cut 1px on mobile view by using that carbon breakpoint.

Indeed, this approach will trim 1px from the image which I thought it's not ok, but I think this was used because the image has a fixed ratio, so trying to use margin or padding won't work if you only want to have that 1px space on mobile. On desktop view it will be ok because using margin-right, for example, it will add the 1px right and bottom.

 I will try to find a cleaner solution 🙂

sstrubberg:
Should we close this PR or put it in draft?

xylish7:
If no one can take a look at it you can put it in draft. At the moment I have a lot of work and can't handle this, but it's pretty important to fix it since is a visual bug.

xylish7:
I've made a new commit. In this one the image won't lose that 1px. Not sure if there is a better solution...Or if it is refactoring the entire component :D, but sadly got not time right now.

packages/gatsby-theme-carbon/src/components/ImageCard/image-card.scss
```diff
@@ -101,6 +99,15 @@
   transition: opacity $duration--fast-01;
 }
 
+.#{$prefix}--image-card__spacing {
```
vpicone, line 102: I feel like this is just as likely to randomly stop working as the original solution haha. If the magic 1px switcheroo is the only solution that's fine, just wanna make sure we've explored more solid options.

PR #977, opened 2020-08-25
fix(FeedbackDialog): update colors and animations

### LINKED ISSUE #790: feedback component: update motion per shixie's specs
## Feedback component: update motion per shixie's specs

Shixie's advice on motion:

- (If the circle is filled) Do an outline of the stroke before filling the circle would help. That’s probably the lower hanging fruit. Draw on the check mark (I was too lazy to do it, but drawing it on vs. fading it in is better)

- dialogue can open with CTA sliding in slightly delayed (also according to guidelines, dialogue should open with x-dir expansion starting before y-dir expansion, but that’s prob too hard to do)

- after user click “submit”, dialogue should close automatically after a brief delay (I used 300ms. I think up to 1 second is all okay) 

- Also shouldn’t we avoid ! according to IBM style guidance? I changed “Thanks” to “Thank you.” 

Shixie made an example and included source files (I added to her file so you can see how the interior stuff animates): https://ibm.box.com/s/hap5dmjj5ltt3ztzh7u017w16d9opli6

In Principle you can see curves and frame timing:

### LINKED ISSUE #886: Feedback button: Make them the same gray color.
currently the feedback button is different color than the jump to top of page button. They should be the same.

—

jnm2377:
Closes #790 
Closes #886 

#### Changelog

- changes Feedback launch button to use same colors as `BackToTop` button.
- changes submitted checkmark animation according to new specs (animate the circle rather than the checkmark)
- changes the animation for `FeedbackDialog`:
    - no longer "fade in" type of animation
    - animates width and height, with width animation being faster/finishing before height
    - button row in feedback dialog is also animated in now (height expansion, too) but with a slight delay (added a delayed state change for this functionality `animateButtonRow`)

**Removed**
- no longer using CSSTransition for the animations


vpicone:
Wicked! Are the buttons supposed to come in at the end? I think things are coming in a little dishevled for the entrance, you can slow it down in the chrome animation tools to help debug it:
Screen Recording 2020-08-25 at 11 43 41 AM

jnm2377:
@vpicone yes to the buttons coming in at the end. Ughhh to the disheveled part. I thought I fixed that. It seems like it happens randomly, then. Need to figure it out... 👀

vpicone:
@jnm2377 The checkmark is gonna be so dope!! At the moment, it gets to about 75% and then snaps to the finish. I wasn't able to get a good gif without pulling the code down an forcing the dialog to stay open but let me know if that tracks.

jnm2377:
@vpicone here's what I'm seeing when I slow down the animation. 
Still trying to figure out how not to let the content wrap, but idk why the button animations are different for you. 
feedback

Also, I think the circle animation appears like it snaps to finish bc of the fill. The circle stroke itself transitions until 99%, and then at 100% I add the bg fill. But I think bc it's happening so fast, it looks like it finishes at 75%.

vpicone:
@jnm2377 huh that's so weird. Is that recording from your local build or the build preview? It seems like they're waiting the 400ms before starting but not sure how to replicate it. I checked ff, chrome, and safari with the build preview and they're coming in delayed.

It looks very cyborgy. Maybe this is a feature and not a bug haha

jnm2377:
@vpicone the buttons **are** supposed to be coming in delayed. But expanding from the bottom up. That was my local build. Also, I updated the checkmark animation. Lmk if that looks better.

vpicone:
@jnm2377 okay cool, the expansion from the middle might just be a side effect of the chrome dev tool  impacting css/js time differently

shixiedesign:
Hey always happy to see motion getting build! I know it's tricky. Here are the main issues I noticed in order of priority: 

- Exit animation of the modal looks not right. I noticed my prototype I didn't include exist. It should first shrink in height, then shrink in width. Currently it's the opposite. If this is too hard to do, consider only shrinking the height. Width change should be secondary & subtle. When the primary motion is width shrinking it feels like a drawer pulling out to the right, which feels very unexpected.

- Buttons sliding up looks a tad tooooo delayed, which is making it feel stuck. The delay should be very small like 10ms-15ms (see screenshot of Principle timeline below)

image

jeanservaas:
Hey @jnm2377 

I'm glad Shixie talked about the easing and motion direction etc. that is def not my specialty. I'd park this guy for a hot sec to finish the color palette page since this is lower priority and animation takes A LOT of noodling. 

A couple of other things to note, Shixie gets at this too, I think the whole animation, panel opening could be sped up in addition to the buttons. 

Also, when you hit the Submit button, the "Thank you" and the check mark icon comes up... after the outline of the checkmark icon draws on, like you've done, the icon needs to fill revealing the knocked out checkmark. This could happen a tad faster too. So if you notice the end state of Shixie's animation in box, this is what you'll end with.

image

jnm2377:
@jeanservaas @shixiedesign I updated the exit animations, and the checkmark animation. One thing to note, it's working fine on Chrome, but the checkmark circle fill isn't animating in Safari. Apparently it's a known bug in safari. I haven't been able to find a workaround. cc: @vpicone have you ever ran into this problem? I tried using `-webkit` prefixes, `%` instead of `to`. It has an initial value, too. But nothing. If I don't animate it, and just give it a permanent fill, it's fine. It just doesn't seem to like the css animation. But the other animated properties are working right, just not fill. There are quite a few threads I found on Stack overflow with this same issue, but nothing helped.

Wondering if we should have a fall back for safari, where the fill is permanent and only the checkmark animates?

vpicone:
This is looking awesome! I'm getting a peculiar bug in Firefox where the top is cut off (Safari and Chrome look fine).
Screen Recording 2020-09-10 at 4 11 02 PM

vpicone (changes requested):
Small Firefox bug

jnm2377:
@vpicone I'm not seeing this bug on ff. I checked both locally and in the preview: https://gatsby-theme-carbon-l6ltgatr9.vercel.app/components/FeedbackDialog

jnm2377:
@vpicone do you have any thoughts on this https://github.com/carbon-design-system/gatsby-theme-carbon/pull/977#issuecomment-686776549 ?
I added a fallback for safari, so the circle is just filled from the start, but I was wondering if you might know of any other solutions.

vpicone:
@jnm2377 SVG animations in safari are a nightmare. Fortunately, it looks like they're getting some love in the [next major version of safari](https://developer.apple.com/documentation/safari-release-notes/safari-14-beta-release-notes#New-Features) coming soon.

Safari accounts for <10% of our user base. If Jean's okay, I'm down for leaning on a temporary fallback and revisiting/hoping for the best in Safari 14.

vpicone:
@jnm2377 so weird. does this help at all? 
Screen Recording 2020-09-11 at 4 36 13 PM

vpicone:
Lookin good what do you think @jeanservaas @shixiedesign ?

vpicone (approved):

sstrubberg (approved):

jeanservaas:
@jnm2377 I asked @johnbister and @shixiedesign to take a look at the timing here!

johnbister:
@jnm2377 it looks good to me!
feebackdialog

jeanservaas:
@vpicone @mjabbink I think we need to revert this. Sorry, I meant for John to offer some suggestions. I feel that the motion/timing is still a little awkward... even though technically it's following our motion guidance. I think that left then up expansion could be a little bit more subtle.

jnm2377:
@jeanservaas if by left and up expansion, you mean width first then height, this is doing exactly that. You can see it in the console screenshot that John posted.

jnm2377:
@jeanservaas we could make the width transition slower or faster, but we can't stagger the transitioning start time because the transitioned properties are on the same element, so the transition will be executed at the same time. As well as the fact that if there is no visible height, you won't notice the width transition at all, so they have to start together.

vpicone:
@jnm2377 should be able to move them independently with key frames right?

```

@keyframes slidein {
  0% {
    margin-left: 100%;
    width: 300%; 
  }

  25% {
    margin-left: 90%;
  }

  75% {
    font-size: 300%;
    margin-left: 25%;
    width: 150%;
  }

  100% {
    margin-left: 0%;
    width: 100%;
  }
`

jnm2377:
@vpicone but if you have no visible height, you won't see the width transition. Does that make sense? That's what was throwing me off. Because even in the video example that Shixie created, they start transitioning at the same time, but the width just finishes way before the height does.

jnm2377:
@vpicone I can open up a PR to try to time it better in keyframes, but they would still have to start at the same time in order for any of it to be visible.

jeanservaas:
@jnm2377 yeah totally understand that it's technically following our guidance — but @vpicone and I went through the same thing when we were working on it initially. Vince seemingly followed the guidance and we just couldn't really make it work... Growing the width and then the height felt clunky and we couldn't smooth it out.

Ultimately, we just threw it out and we agreed that Vince would basically just copy the motion on the Gatsby site's feedback component. Shixie gave us some advice for the motion and I added it to the issue thinking that maybe it would shed some light on how to make this work correctly.

I think before moving forward with this, just like John's PRs, we're going to need a review from one of the system's motion experts. @shixiedesign is supposed to weigh in on this tonight or tomorrow. If she can't get to it I will try to have Wonil or even Pete Garvin review it.

mjabbink:
Agree it needs a bit more work

johnbister:
dialogmock

Would something like this be a little closer?

shixiedesign:
Hey everyone. It actually looked much better from the last time I saw it but I'd agree with not merging. It's still calling too much attention to itself. 

1. firstly let's lock / freeze what's NOT supposed to move. Everything inside the module minus the buttons shouldn't move relative to itself. Think of it as an image inside the parent container that gets masked when the parent shrinks/grows. Right now we catch moments like this where the little emojis are all squished 

image

2. I see the width-then-height motion is there in the entrance. Well done! It's probably clunky because it is taking too long. In my prototype, the entire module reaches full width and height by `240ms`. It took 400ms in the build.

3. Buttons are still not the last to finish. The buttons should arrive at final position after module reaches full height.
image

4. 👍 closing animation is much better than before. It lacks the width-then-height thing but it's okay in this case. Much more efficient. 

Here's a box link to the video prototype. I improved it a bit (color and added placeholder box for "content") but motion is unchanged. https://ibm.box.com/s/2o0eacmb4knebvpnba5vjldzk4sjhb0r

mjabbink:
I think this works better @shixiedesign.

johnbister:
dialogmockup

I was able to move around the existing keyframes you had and get it pretty close to Shixie's prototype.  Let me know if this is helpful.

mjabbink:
@shixiedesign @johnbister Maybe a bit slow?

mjabbink:
Is closing a hair too fast? Maybe fine @shixiedesign @johnbister

mjabbink:
@jeanservaas @jnm2377 The colors in preview are not right. I think on active state it should remain as a hover and not be dark and a solid face. @jeanservaas can you revisit this and provide spec.

mjabbink:
There is a lot of color/hover activity for such a simple thing. 




Screen Shot 2020-09-16 at 9 53 49 AM

jeanservaas:
@jnm2377 

Enabled state should look like this:

image


Both hover and selected state should look like this:

image

mjabbink:
@jnm2377 Seems more simple. @jeanservaas whatta ya think?

vpicone:
@jeanservaas @mjabbink reverting this PR, please make an issue with a proper motion spec.

johnbister:
> @shixiedesign @johnbister Maybe a bit slow?

@mjabbink I sped it up a bit.  Let me know if you think this works as a compromise.  Any faster and I have a hard time seeing any easing out.  
[dialog_faster.mp4.zip](https://github.com/carbon-design-system/gatsby-theme-carbon/files/5233874/dialog_faster.mp4.zip)

PR #991, opened 2020-09-10
fix: ordered list alignment in safari

### LINKED ISSUE #985: List should not be hanging into gutter
via @mjabbink 

- [x]  ordered list numbers should left align with body copy. Not hang into gutter

jnm2377:
Closes #985 
 
- adds webkit-padding for safari only
- **testing**: view in Safari browser, compare localhost/preview to existing list example on live site and see alignment has changed: https://gatsby-theme-carbon.now.sh/components/markdown#lists


packages/gatsby-theme-carbon/src/components/markdown/Markdown.module.scss
```diff
@@ -85,6 +85,7 @@ ul :global(.bx--list--unordered) {
 
 :global(.bx--list--ordered:not(.bx--list--nested)) {
   margin-left: 1ch;
+  -webkit-padding-start: 0.5ch; // needed to align ordered lists to body copy in Safari
```
vpicone, line 88:
```suggestion
  -webkit-padding-start: 0.5ch; // needed to align ordered lists to body copy in Safari #985
```

vpicone (approved):
Awesome, adding the issue numbers can be helpful I think, especially now that VS Code let's you open them from the editor

vpicone (changes requested):
IMG_F58B955E6D95-1

It looks like the nested list items could use this as well, maybe update the rule at line 75 to be:

```
:global(.bx--list--ordered) {
  list-style: decimal;
  -webkit-padding-start: 0.5ch; // needed to align ordered lists to body copy in Safari #985
}
```

Not sure why the `, ol` was there but looks like it can be removed.

jnm2377:
@vpicone looks like the nested list alignment was off on chrome too, so I updated it for both. I also changed the way I was targeting the safari only fixes bc for some reason it was affecting chrome too. Should be good now.

packages/gatsby-theme-carbon/src/components/markdown/Markdown.module.scss
```diff
@@ -106,6 +106,21 @@ ul :global(.bx--list--unordered) {
   margin-left: 1ch;
 }
 
+:global(.bx--list--ordered.bx--list--nested) {
```
vpicone, line 109: @jnm2377 Do you mind maybe moving the comment up hear and making it a little more descriptive? I think this markdown style sheet is getting buck wild and I won't remember what happened in a month haha.

vpicone (approved):

PR #995, opened 2020-09-14
Permit overwrite of MdxNode field addition to page context

mikyjpeg:
The block of code doesn't work in conjunction with some customization (e.g. when `page.component` is defined to point to a template or when pages are not in `src/pages`. Adding a blog to avoid re-running the block of code if the MdxNode has been already defined will permit to create additional `onCreatePage` method to apply custom logic.

#### Changelog

**Changed**

- gatsby-node.js

### DIFF
--- a/packages/gatsby-theme-carbon/gatsby-node.js
+++ b/packages/gatsby-theme-carbon/gatsby-node.js
@@ -25,24 +25,26 @@ exports.onCreatePage = (
   { page, actions, getNodesByType, ...rest },
   pluginOptions
 ) => {
-  // Find the MdxNode that created our page
-  const MdxNode = getNodesByType('Mdx').find(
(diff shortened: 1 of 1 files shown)


packages/gatsby-theme-carbon/gatsby-node.js
```diff
@@ -25,24 +25,26 @@ exports.onCreatePage = (
   { page, actions, getNodesByType, ...rest },
   pluginOptions
 ) => {
-  // Find the MdxNode that created our page
-  const MdxNode = getNodesByType('Mdx').find(
-    ({ fileAbsolutePath }) => fileAbsolutePath === page.component
-  );
+  if(!page.context.MdxNode){ // Enable the possibility to overwrite it
```
vpicone, line 28:
```suggestion
  // Don't override if it's already been provided
  if(!page.context.MdxNode){
```

PR #1000, opened 2020-09-17
Update _plex.scss

### LINKED ISSUE #927: [plex] font-weights missing, only 400 and 600 available
## Detailed description

We only have two font weights available on our site that we generated from the starter a few months ago. We don't have any font weight overrides in place in our styles.

- [ ] 300 font weight is not included by default - it appears we only have 400 and 600.
- [ ] symptom of the above issue is that the page headers on our site are not the proper weight ([Internal issue here](https://github.ibm.com/ai-applications/design/issues/149))
- [ ] `additionalFontWeights` config doesn't include the specified weights either
  - we have `additionalFontWeights: ['200', '200i'],` defined, but as you can see in the gif below demonstrating all these issues, we only have the two weights.

Jul-20-2020 09-41-52

We're on v1.24.3, but I don't think an upgrade will fix. I didn't see any updates that have been made since then up until current.

vpicone:
closes #927

### DIFF
--- a/packages/gatsby-theme-carbon/src/styles/internal/_global.scss
+++ b/packages/gatsby-theme-carbon/src/styles/internal/_global.scss
@@ -21,6 +21,7 @@ $feature-flags: (
 @import 'website-side-nav';
 
 // Typography
+@include carbon--type-reset();
 @import '~@carbon/elements/scss/type/classes';
 @import 'plex';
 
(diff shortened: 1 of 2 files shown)


packages/gatsby-theme-carbon/src/styles/internal/_global.scss
```diff
@@ -25,10 +25,6 @@ $feature-flags: (
 @import '~@carbon/elements/scss/type/classes';
 @import 'plex';
 
-$font-family: 'IBM Plex Sans', 'Helvetica Neue', Arial, sans-serif;
```
vpicone, line 28: These are wrong, but aren't actually doing anything

PR #1002, opened 2020-09-22
fix: switcher aria label text

### LINKED ISSUE #847: The word "navigation" in "IBM Design ecosystem navigation" is redundant
See this line of code, where a `<nav>` element has `aria-label="IBM Design ecosystem navigation"`:
https://github.com/carbon-design-system/gatsby-theme-carbon/blob/497dca3c2e01f514ab87ead50967da2825461dd3/packages/gatsby-theme-carbon/src/components/Switcher/Switcher.js#L12

The word "navigation" should just be deleted from the aria-label string, because screen readers already say "navigation" for a nav element, so it would sound like:

> "IBM Design ecosystem navigation navigation"

jnm2377:
Closes #847

- updating aria label so it won't be redundant for screen reader


vpicone (approved):
🚀

PR #1003, opened 2020-09-22
fix(PageTabs): update current tab regex 

### LINKED ISSUE #884: tab navigation is broken when section title contains tab title
## Detailed description
On https://pages.github.ibm.com/ai-applications/design/conceptual-model/model

the `learn more` and `overview` tabs are broken: 

It’s generating urls of
.../design/conceptual-learn-more/model
but it should be
.../design/conceptual-model/learn-more

so it 404s

### LINKED ISSUE #960: Problem with a tab name that has an apostrophe in it
## Detailed description

> Describe in detail the issue you’re having.
tabs:
  [
    'Overview',
    'Welcome page',
    'Whats new',
  ]
Results in an error on the page and the tab list getting displayed as plain text

Using the HTML code for ' instead of the character results in the tab name being displayed as What&#39; New

I would like to be able to use an apostrophe in the tab name.

Firefox

1. Create a page that has a tab that as an apostrophe in the name.

## Additional information

- Screenshots or code
- Notes

jnm2377:
Closes #884 
Closes #960 

- fixes current tab regex so it's not replacing incomplete slugs
- removes apostrophes from tabs/slugs 
- adds documentation for tabs with `'`
- testing example added under `> Guides > Testing tabs` and `> Guides > Navigation > Nav`

**Note: Do not merge after approval, need to remove testing pages before merging**


packages/gatsby-theme-carbon/src/components/PageTabs/PageTabs.js
```diff
@@ -15,15 +15,21 @@ import {
 export default class PageTabs extends React.Component {
   render() {
     const { tabs, slug } = this.props;
-    const currentTab = slug.split('/').filter(Boolean).slice(-1)[0];
+    const newSlug = slug.replace(/[’']/g, ''); // removes any apostrophes from slug
```
vpicone, line 18: While this solves the apostrophe problem, we probably wanna look for a solution that works for any punctuation/odd characters

packages/gatsby-theme-carbon/src/components/PageTabs/PageTabs.js
```diff
@@ -15,15 +15,21 @@ import {
...
+    const currentTab = newSlug.split('/').filter(Boolean).slice(-1)[0];
 
     const pageTabs = tabs.map((tab) => {
-      const slugifiedTab = slugify(tab, { lower: true });
+      const tempTab = tab.replace(/[’']/g, ''); // removes any apostrophes from tab
+      const slugifiedTab = slugify(tempTab, { lower: true });
```
vpicone, line 23: Slugify has a [`strict` option](https://www.npmjs.com/package/slugify#options) that removes any punctuation/special characters

From a quick test, it strips out the apostrophe so it would match the yaml/filename after processing. I think that should remove the need for the temp tab all together maybe.

packages/gatsby-theme-carbon/src/components/PageTabs/PageTabs.js
```diff
@@ -15,15 +15,21 @@ import {
...
       const selected = slugifiedTab === currentTab;
-      const currentTabRegex = new RegExp(`${currentTab}(?!-)`);
+      const currentTabRegex = new RegExp(`(?<!-)${currentTab}(?!-)`);
```
vpicone, line 25: While this works for `/conceptual-model/model` it wouldn't solve the issue for something like `/navigation/nav`. We'd still end up replacing the first instance of `nav` and get broken urls.

What if instead, we use a `$` to indicate the end of the string (while allowing for an optional trailing slash)?

```suggestion
      const currentTabRegex = new RegExp(`${currentTab}/?$`);
```

vpicone (changes requested):

packages/gatsby-theme-carbon/src/components/PageTabs/PageTabs.js
jnm2377: Good point @vpicone. Would it still work if the url is pointing to a specific section of the page and the tab isn't necessarily the end of the string? i.e. `/conceptual-model/model/#section-of-page`

vpicone: @jnm2377 good question, hash links don't work if there's a slash in between though ( so it'd be `/conceptual-model/model#section-of-page` and `/conceptual-model/model#section-of-page/`) I'm not even sure they work with a trailing slash.

Let me know if it doesn't work with section links and we can brainstorm

jnm2377: this example I was just looking at has a slash in between: https://www.ibm.com/design/language/iconography/pictograms/contribute/#export-svgs-from-adobe-illustrator

vpicone: @jnm2377 oh wow sure enough maybe this is a server setting? Good catch. We can get around this by looking for _either_ a hashtag _or_ the end of the line. I'll edit my suggestion.

vpicone: @jnm2377 updated to add an optional group for hash links

jnm2377: Seems to be working. I added another testing example under `Guides > Navigation > Nav`

packages/gatsby-theme-carbon/src/components/PageTabs/PageTabs.js
jnm2377: Hmm.. @vpicone we can add more characters into this regex. But wondering if there would be an easier way to say "replace any character that's not a letter, number or `-`" besides listing all the possible characters.. I'll look into it. Regex's aren't my strong suit but I'll google it haha

jnm2377: updated!

packages/gatsby-theme-carbon/src/components/PageTabs/PageTabs.js
```diff
@@ -15,12 +15,15 @@ import {
 export default class PageTabs extends React.Component {
   render() {
     const { tabs, slug } = this.props;
-    const currentTab = slug.split('/').filter(Boolean).slice(-1)[0];
+    const newSlug = slug.replace(/[^a-zA-Z0-9-\/]/g, ''); // removes any characters that aren't letters, numbers or dash/fwd slash
```
vpicone, line 18: Do we still need this? If so, we could use slugify w/ the same settings to remove the stuff so we can guarantee we're comparing apples to apples.

vpicone (commented):
qq

packages/gatsby-theme-carbon/src/components/PageTabs/PageTabs.js
jnm2377: So slugify the slug? This part is for the initial landing tab, the line where we slugify the tab name is just for when we're looking for what to replace when switching tabs, so we still need something to replace the characters on the initial tab. Are you suggesting something like:
`const newSlug = slugify(slug, { lower: true, strict: true });`

jnm2377: ^ hope that makes sense regarding the slug v tab regex @vpicone

vpicone: @jnm2377 the slug comes from `location.pathname` in the Default template so it should already be a valid URL. After the update to use strict slugify below, I'm not sure what links they would click that could take them to a URL with special characters in it.

vpicone (approved):
🚀

sstrubberg (approved):
That's some legit regex'ing 🎉

PR #1013, opened 2020-10-01
fix: replace orange with blue

### LINKED ISSUE #986: fix[component]: gatsby code snippet showing wrong color theme
Not sure how this happened, but the Gatsby code snippet is showing orange type. @tw15egan working on a color theme for the Gatsby snippets a long time ago and it's been the same for a while, it there were teals, purples, blues and magentas but no orange. It was so long ago that I don't have specs for it, it was something we worked through together and PR'd.

**Wrong:**

image

You can see remnants of the old code snippet color theme on other pages. I'm wondering what the best way is for us to revert back to this. 

These are the original theme colors:

image

**IDL:**
image

**Carbon:**
image

jnm2377:
Closes #986 

- replaces orange with blue


jeanservaas (approved):

sstrubberg:
I've actually been going in a replacing all of those with neutral values for the Theme'ing PR.

[Here is an example of dark mode](https://dark-dark.netlify.app/components/code-blocks).

[And here is an example of white mode](https://white-white.netlify.app/components/code-blocks/).

Should I be looking to add these little color moments for both light and dark themes?

vpicone:
@sstrubberg yeah we can might just want to incorporate these tweaks into your theming PR

jnm2377:
I'm fine with adding them to your PR. I added the comments bc I thought they were helpful since they don't use tokens or variables. @sstrubberg

sstrubberg:
I'd love it if we could all sync up on a call regarding this. I have a few questions that would be cumbersome in a github thread.

PR #1015, opened 2020-10-03
chore: hiding autolink header on mobile

### LINKED ISSUE #605: Header link off screen and undiscoverable on mobile
It's really hard to find these links on mobile, since there's no hover effect. They're also off screen some.

Screen Shot 2019-12-11 at 1 07 34 PM

luizamendes:
Closes #605

Changing position of AutolinkHeader on mobile devices

#### Changelog

**Changed**

- Now the Autolink Header appears on the right on mobile devices.
- No changes were made on the desktop current behavior

This is how it looks on the hovering state:
autolink-header-on-mobile-hove


vpicone:
Hey @luizamendes thanks for the PR!! I’m sure this is the best approach, we still want folks to be able to find the link and use it on mobile. Also this makes for a worse accessibility experience for screen readers on mobile. It’d be better just to not render it at all.

luizamendes:
@vpicone Thanks so much for the feedback. I closed the PR as it seems to still have a discussion going on about this issue. I'll be glad to come back to this after a decision has been made.

vpicone:
@luizamendes what about making an anchor _after_ the header that's only visible on mobile hover/tap

luizamendes:
@vpicone By doing that, we would have to remove the position `absolute` from the anchor, which would make the title move to the right on desktop, and on mobile the anchor could break to the next line... Was that what you thought?

vpicone:
@luizamendes would you have to remove it? For the one on the left you could give it `visibility: hidden` And it wouldn’t move. 

White-space: no wrap could prevent wrapping.

luizamendes:
@vpicone I have updated the PR. Set the `position: absolute` for the desktop only. Now the anchor icon is on the right on mobile devices.  I attached an image of how it looks.

vpicone:
@luizamendes BOOM love it

vpicone (approved):
Great use of the useMedia hook! It’s not great for things that are rendered on the initial page load (all the js has to load for it to have a proper value) but since these are hidden it’s great.

luizamendes:
@vpicone Thank you! :) Just out of curiosity and for future contributions, what approach would you use for the initial page load? What file has this approach, so I can take a look at it? Thanks!

vpicone:
@luizamendes thats a good question, it would depend on the sort of thing we're rendering. `useLayoutEffect` is good if you need to measure something before it's rendered. I might also look at one of the `gatsbyBrowser` or `gatsbySSR` to handle how the initial html is rendered.

PR #1029, opened 2020-10-09
chore: ensure we're passing through className props to allow for easier style overrides

vpicone:
Realized how dumb it was we can't pass classNames into things like `PageHeader`, I'm mainly concerned about components folks use in markdown.


jnm2377 (approved):

packages/gatsby-theme-carbon/src/components/Aside/Aside.js
```diff
@@ -1,29 +1,15 @@
 import React from 'react';
-import PropTypes from 'prop-types';
 import classnames from 'classnames';
```
sstrubberg, line 2: Sometime we are importing as `classnames`, but most of the time we import as `cx`. Should we standardize on just using `cx` for consistency? Or does that not really matter? Also, *this is awesome*!

vpicone: Yeah we totally should, I'll double check this

vpicone: @sstrubberg went through and normalized everything towards cx. It might conflict with some of your stuff so we'll merge your PR first

sstrubberg:
@vpicone I think we're ready to bring this in once the conflicts are resolved.

sstrubberg (approved):

PR #1034, opened 2020-10-19
fix(FeatureCard): image size for larger breakpoints

### LINKED ISSUE #1033: fix(FeatureCard): Breaks down on XLG screens using HeaderNav navigation style.
When using HeaderNav navigation style the FeatureCard starts to break down at some point during LG screens and completely breaks down at XLG screens.

https://w3.ibm.com/design/racial-equity-in-design/our-work/

image

sstrubberg:
Closes #1033 

When using the `HeaderNav` navigation style, the FeatureCard breaks down at larger breakpoints. I overrode the image size set by gatsby to the CSS default which is `none`.

** DO NOT MERGE UNTIL I REMOVE THE NAVIGATION STYLE **


packages/gatsby-theme-carbon/src/components/FeatureCard/feature-card.scss
```diff
@@ -89,4 +89,10 @@
 .#{$prefix}--feature-card .gatsby-resp-image-wrapper {
   margin-left: 0;
   margin-right: 0;
+  @include carbon--breakpoint('lg') {
+    max-width: 100% !important;
```
jnm2377, line 93: does it not work without the `!important`?

sstrubberg: I wish it did. But it was the only way I could get it to take precedence over gatsby's insistence that the max-width be hardcoded at 1152px.

sstrubberg:
@vpicone @jnm2377 I changed the max-width to `none`. Curious on your thoughts.

vpicone:
> @vpicone @jnm2377 I changed the max-width to `none`. Curious on your thoughts.

I don't really know the difference and it bothers me haha

packages/gatsby-theme-carbon/src/components/FeatureCard/feature-card.scss
```diff
@@ -68,7 +68,7 @@
     display: block;
     margin-left: -1rem;
     margin-right: -1rem;
-    max-width: 100%;
+    max-width: none !important;
```
vpicone, line 71: did we still need important?

vpicone (commented):
qq

packages/gatsby-theme-carbon/src/components/FeatureCard/feature-card.scss
sstrubberg: unfortunately, yes.

jnm2377:
What should we be doing with this PR? cc: @vpicone @sstrubberg I'm wary of shipping any styles with `!important` to production, but also not sure how big of an issue this is. Should we close this for now until we find another solution?

sstrubberg:
> What should we be doing with this PR? cc: @vpicone @sstrubberg I'm wary of shipping any styles with `!important` to production, but also not sure how big of an issue this is. Should we close this for now until we find another solution?

Without it, FeatureCards are mega-busted using HeaderNav navigation style.

jnm2377:
@sstrubberg I worked on dropdown for the initial v10 release, so I'm happy to pair up if you want a hand figuring out the styles.

jnm2377 (approved):
hmm.. if this is the only solution, then I suppose it's ok. I would add a comment about why it's needed (linking the issue #) and then maybe a to-do: to see if we can find another solution in the future that doesn't require `!important`

vpicone:
@sstrubberg I'm still getting feature card weirdness
Screen Shot 2020-11-23 at 2 17 23 PM
Screen Shot 2020-11-23 at 2 17 02 PM

sstrubberg:
Ok, having had the time to reflect on this PR, I think I want to close it, but hear me out first...

I think this problem is not actually a FeatureCard problem but a problem with the fact that GTC is using the 12 column grid and not 16. The FeatureCard is configured in such a way that it fits perfectly on any GTC page, but when you eliminate the leftnav and all the width that comes with it, it screws with the FeatureCard and throws off the ResourceCard. Thoughts?

vpicone:
@sstrubberg I'd have to do more digging, that could be. These problems only exist in header navigations style? If we can come up with a reproducible case, I agree we should increase the scope of the issue.

jnm2377:
I think you might be right about it being a grid thing @sstrubberg . Maybe we need to use 16 grid if it's configured for a header only, not sure how we would access that header query from a scss file though.... 🤔 Have you tried using 16 col grid locally to see if it fixes it?

jnm2377:
Should we close this? @sstrubberg

sstrubberg:
I think once the REiD site goes public and if we ever re-create ibm.com/design using the headernav style, we're going to be confronted with this again.. Until we meet again 12col grid!

PR #1044, opened 2020-10-29
style(headernav): active state

sstrubberg:
@jnm2377 and I just had a glorious pair-programming session and fixed up the header when using the headernav navigation style. Previously, there was no blue bar on any item to indicate where you were on the site. Now you do!

Top level items:
image

Menus with sub-items that have been selected:
image

and the sub-items themselves (now with rules!):
image


jnm2377 (approved):
🎉

jeanservaas (commented):
Hey guys, 

The selected state for the dropdown would actually not have the blue bar in this instance, instead the selected item background would change to $selected-ui, see below.

image

jeanservaas (approved):
boom. thank you!

packages/gatsby-theme-carbon/src/components/Header/Header.module.scss
```diff
@@ -1,5 +1,39 @@
...
+  background-color: $carbon--gray-80;
+}
+
+// Removes the blue line that sits under an active item in menu.
+.isActive :global(.bx--header__menu-item--current)::after {
+  display: none;
+}
+
+// Overrides the style from upstream Carbon to give a blue60 instead of a blue50
+.isCurrentPageOverride :global(.bx--header__menu-item--current)::after {
```
vpicone, line 26: whats the diff between is active and iscurrentpage?

packages/gatsby-theme-carbon/src/components/Header/Header.module.scss
```diff
@@ -1,5 +1,39 @@
 @import '~carbon-components/scss/globals/scss/layout';
 
+// Creates the blue line under the active menu item
+.isActive::after {
+  position: absolute;
+  top: 0;
+  right: 0;
+  bottom: -2px;
```
vpicone, line 8: should this be -3?

packages/gatsby-theme-carbon/src/components/Header/Header.module.scss
sstrubberg: I made better descriptors of these selectors. Hopefully, they are easier to understand.

packages/gatsby-theme-carbon/src/components/Header/Header.module.scss
sstrubberg: Whoops! This actually needs to be 0

vpicone:
@jeanservaas should the active tab (or all the tabs?) be using white for the text color? IBM Design and the page level tabs use white for all of the tabs.

Also, the Page tabs are using 3px, this is using a chonkier 4. Do we want to settle on one or the other?

### Page Tabs
Screen Shot 2020-11-09 at 3 52 21 PM

### This PR
Screen Shot 2020-11-09 at 3 54 38 PM

jeanservaas (approved):

jeanservaas:
@vince good catch. That type should be white, always (not just one selected) if we're using ibm.com/design and the accessibility as the standard.

image

image

image

Also, with the tab chunkiness and inconsistency, there is a bigger issue and I didn't want to hold up this PR to deal with it. 

There is inconsistency with tab bar chunkiness between us and .com team that I wanted to bring up in the next big Carbon meeting... so we can decided once and for all what we want the standard to be.

image

jeanservaas (commented):
@vince made a good point, header text should be white on enabled and stay white in all interactive states — I know that's not how our side nav behaves but it's the standard top nav behavior we set for ibm.com/design and accessibility sites.

vpicone:
@jeanservaas should we make it 3px so we have consistent chonk within the theme? 

We can always change them together (or leave them) after the team convo.

jeanservaas (commented):
@vince and @sstrubberg sorry for all the confusion. Just had the Big Carbon call with Mike / Sadek and .com and some decisions were made.

- [ ]  We **DO** want to change the bar to **3px** tall on the top nav. That IS what .com team is using and that's what we use on tabs and on the Do / Don't components

- [ ] .com team and Mike agreed that we want the **enabled type** on the top nav on the top nav to be **$text-02 token** (g30) the **hover** and the **selected** **/ active states** will use the **$text-01 token** (white)

- [ ] We also universally decided that we want the top nav items to remain 16px as a default ($body-short-02) — Mike originally wanted to go to 14px but we convinced him otherwise. 

- [ ] Can we update ibm.com/design and https://www.ibm.com/able/ and https://www.ibm.com/impact/be-equal/ to use this component as well once it's merged? Would that be another issue?

jeanservaas (commented):
ug sorry @jnm2377 I was trying to get a COMMENT icon instead of an approval next to my name and I accidentally clicked and removed your approval...ugggg

vince:
@jeanservaas You are summoning the wrong vince. Please be careful. I think everyone on this project has made this mistake and it's getting a bit aggravating.

vpicone (changes requested):

jeanservaas:
Looking good! One thing missing is just the rules from the dropdown, sorry I didn't catch it before:

**Carbon dropdown:**
image


Rule spec:

Weight: 1px stroke
Color: $ui-03
Margin: 16px on either side

The rules disappear around the field that has hover or focus.

sstrubberg:
> Looking good! One thing missing is just the rules from the dropdown, sorry I didn't catch it before:
> 
> **Carbon dropdown:**
> image
> 
> Rule spec:
> 
> Weight: 1px stroke
> Color: $ui-03
> Margin: 16px on either side
> 
> The rules disappear around the field that has hover or focus.

Got everything humming along just fine. For the life of me, I can't figure out this little border trick. The hover-states are making is super tricky. I left a bunch of commented out code to illustrate the direction I was heading. @vpicone after turkey-break, I'd love to jam with you on this.

jeanservaas:
@sstrubberg 

Yes the rules disappear on the field that hovers and we've had problems with that, especially with accordion, i remember. Sorry that it's a hassle, I know some of our system devs have also had to figure this out, like maybe andrew and tj.

jnm2377:
@jeanservaas if by dropdown you mean the header menu items, then we should open up a design issue for this upstream. We can add a temporary fix here, but we're using carbon components for the header and if that's the style that should be applied, it's not being applied there.

Screen Shot 2020-11-23 at 10 11 51 AM
Screen Shot 2020-11-23 at 10 12 49 AM

jnm2377:
@sstrubberg I worked on dropdown for the initial v10 release, so I'm happy to pair up if you want a hand figuring out the styles.

jeanservaas:
Hey guy, I just wanted to address an issue that came up in  Carbon-support: 
https://ibm-studios.slack.com/archives/GD8JG6JR3/p1606854462005300

1. The Gatsby shell has gotten ahead of the upstream Carbon shell in many ways because Aux has had to move a lot faster than systems

2. The rules in the shell dropdown in Gatsby have been present since /Design was built (going on two years). 

3. Yes it would be ideal to do upstream shell work in Carbon... will that happen soon, probably not. But agreed that we should make an issue and spec this stuff there as well so everything is in sync. @aagonzales can probably speak to other shell issues that have not been dealt with yet... unless I'm wrong, I'm not sure we've even addressed theming in the shell.

4. Since we've already go these rules and various other header overrides existing in Gatsby, it makes sense to streamline the.

sstrubberg:
> Hey guy, I just wanted to address an issue that came up in Carbon-support:
> https://ibm-studios.slack.com/archives/GD8JG6JR3/p1606854462005300
> 
> 1. The Gatsby shell has gotten ahead of the upstream Carbon shell in many ways because Aux has had to move a lot faster than systems
> 2. The rules in the shell dropdown in Gatsby have been present since /Design was built (going on two years).
> 3. Yes it would be ideal to do upstream shell work in Carbon... will that happen soon, probably not. But agreed that we should make an issue and spec this stuff there as well so everything is in sync. @aagonzales can probably speak to other shell issues that have not been dealt with yet... unless I'm wrong, I'm not sure we've even addressed theming in the shell.
> 4. Since we've already go these rules and various other header overrides existing in Gatsby, it makes sense to streamline the.

Thx for the tip on ibm.com/design. I'll spend some time dissecting that and see if that bears fruit.

packages/gatsby-theme-carbon/src/components/Aside/Aside.js
```diff
@@ -15,6 +15,8 @@ const Aside = ({ children, className, hideRule, ...rest }) => {
   );
 };
 
+export default Aside;
+
```
jnm2377, line 19: why is this here if it's already being exported below?

packages/gatsby-theme-carbon/src/components/markdown/P.js
```diff
@@ -19,6 +19,8 @@ const P = ({ children, className, fullWidth, ...rest }) => {
   );
 };
 
+export default P;
+
```
jnm2377, line 23: also already being exported below.

jnm2377 (changes requested):
Aside and P have double exports

packages/gatsby-theme-carbon/src/components/Aside/Aside.js
sstrubberg: blergh.. at one point they were broken so i fixed them... then someone else fixed them on a different PR and merged to main.. then i merged from main... now i have two! mergemergemerge! :P

jeanservaas (approved):
Thanks Scott!

packages/gatsby-theme-carbon/src/components/Header/Header.module.scss
```diff
@@ -121,3 +192,37 @@ header.header {
     border-left: 1px solid $carbon--white-0;
   }
 }
+
+.header :global(a.bx--header__menu-item:hover) {
+  color: $carbon--white-0;
+}
+
+// Style for the subtle border under each item in the headernav sub menu.
+.header
```
vpicone, line 201: There's a lot of selectors here that make it tough what the actual intent is. Do we need to use each selector along the way? Why not just `.header .bx--header__menu-item` or better yet, why not a class that we throw onto the menu item ourselves?

packages/gatsby-theme-carbon/src/components/Header/Header.module.scss
```diff
@@ -121,3 +192,37 @@ header.header {
     border-left: 1px solid $carbon--white-0;
   }
 }
+
+.header :global(a.bx--header__menu-item:hover) {
+  color: $carbon--white-0;
+}
+
+// Style for the subtle border under each item in the headernav sub menu.
```
vpicone, line 200: In general we want comments to explain **why** something was done a certain way. If you find yourself explaining **what** is happening, there's a good chance the code could be clearer.

packages/gatsby-theme-carbon/src/components/Header/Header.module.scss
```diff
@@ -121,3 +192,37 @@ header.header {
...
+    border-bottom: 1px solid $carbon--gray-80;
+  }
+
+  &:hover::after {
+    display: none;
+  }
+}
+
+// Removes the last subtle border in a menu from creeping onto the page.
+.header
```
vpicone, line 222: Same as above

vpicone (changes requested):

packages/gatsby-theme-carbon/src/components/Header/Header.module.scss
```diff
@@ -22,23 +11,70 @@ header.header {
...
+  outline-offset: -2px;
+}
+
+.header button {
+  display: flex;
+  justify-content: center;
+  align-items: center;
+}
+
+header.header {
```
vpicone, line 69: Did you need the extra specificity here? If so, just make a note of that so someone doesn't just assume they can remove it.

vpicone (approved):
Good to go once we get clarification on that `header.header` Thanks for being so patient. I think you'll agree it looks a good bit more readable now

jnm2377 (approved):
Looks great!!!! 🎉 🎉 🎉 🎉

PR #1050, opened 2020-11-02
fix(search): update search styles

### LINKED ISSUE #1049: [search] extend search field in mobile to allow users to enter a term
1. Right now in mobile, it's not possible for users to trigger the typewriter to enter a search term on their phones because the field does not extend wide enough.

2. Mike wanted to try to get the search width onto the grid at the desktop and tablet breakpoints as well. 

### Current

IMG_2256

Please span the search to the left side of the hamburger menu (see below) to make it as easy as possible for users to interact with it.

### Expected

image

## 2. Modify search width to align with grid at all breakpoints.

Align edge of open Search field with the edge of the first tab on the banner at all breakpoints.

### Expected

1312 breakpoint example (do same alignment through tablet)
image

alisonjoseph:
Closes #1049 

#### Changelog

- Updated styles for search at all breakpoints. Should align on grid at desktop, and span full width at tablet/mobile.


sstrubberg (approved):
💥

jeanservaas (approved):
Wow quick turnaround! Looks good to me.

vpicone (approved):

mjabbink:
Way nicer and especially in mobile! Look forward to this puppy hitting the glass.

mjabbink:
@alisonjoseph and @vpicone The above Preview link is what I was responding to in this issue and it’s not what is live. So Let’s get on a 5-minute call to clear up because I’m missing why we did not deploy this preview

Issue #107, opened 2019-05-25
ArticleCard: Add two typographic situations to ArticleCards


mjabbink:
@alisonjoseph and @vpicone 

Expected:
The below type hierarchies should be added to the ArticleCards. Currently these exists within ResourceCards which is good but this same usage/style exists in ArticleCards too underneath images.

<img width="798" alt="Screen Shot 2019-05-25 at 12 30 15 PM" src="https://user-images.githubusercontent.com/32881239/58372247-f33ec380-7ee8-11e9-94d5-8c93c9dda4f5.png">

——

Using the type specs from ResourceCards below:

<img width="374" alt="Screen Shot 2019-05-25 at 12 22 01 PM" src="https://user-images.githubusercontent.com/32881239/58372262-123d5580-7ee9-11e9-87a9-2f4d3cfeb239.png">

——

<img width="391" alt="Screen Shot 2019-05-25 at 12 22 06 PM" src="https://user-images.githubusercontent.com/32881239/58372270-26815280-7ee9-11e9-8045-00c3a1df5538.png">

PR #1056, opened 2020-11-11
fix(switcher): added reid and universal experiences sites

sstrubberg:
Closes [#107](https://github.ibm.com/Design/racial-equity-in-design/issues/107)

Added Racial Equity in Design and the Universal Experiences sites to the Switcher.

DON'T MERGE UNTIL REiD GOES PUBLIC

### DIFF
--- a/packages/gatsby-theme-carbon/src/components/Switcher/Switcher.js
+++ b/packages/gatsby-theme-carbon/src/components/Switcher/Switcher.js
@@ -103,7 +103,7 @@ const DefaultChildren = () => (
       Carbon Design System
     </SwitcherLink>
     <SwitcherLink href="https://www.ibm.com/standards/web/carbon-for-ibm-dotcom/">
-      Carbon for IBM.com 
+      Carbon for IBM.com
     </SwitcherLink>
     <SwitcherLink href="https://www.ibm.com/design/event/">
       IBM Event Design
@@ -127,6 +127,14 @@ const DefaultChildren = () => (
     <SwitcherDivider>Community</SwitcherDivider>
     <SwitcherLink href="https://w3.ibm.com/design/" isInternal>
       IBM Design
+    </SwitcherLink>
+    <SwitcherLink href="https://w3.ibm.com/design/careerplaybook" isInternal>
+      IBM Design Career Playbook
+    </SwitcherLink>
+    <SwitcherLink
+      href="https://w3.ibm.com/design/racial-equity-in-design"
+      isInternal>
+      Racial Equity in Design
     </SwitcherLink>
   </>
 );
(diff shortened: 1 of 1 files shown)


jnm2377:
just wondering if this was approved by @mjabbink?

mjabbink:
I don’t know about the playbook. I’ve never seen that site or knew it existed.

The REiD should not have a lock. It will be public on launch

mjabbink:
someone send me link to the playbook site. Is it a Gatsby site?

mjabbink:
@sstrubberg The switcher is getting loaded up now and noticed it does not scroll if browser winder is made shorter it crops and no way to get the bottoms items. Should I make separate issue to add scrolling to switcher?

mjabbink:
Also not so convinced the playbook should be in the switcher. It seems to have good visibility in w3/design. I can actually image this content inside that site actually. Not sure why this needs to be a separate site? The switcher is high level entry point to IBM design and we cannot include everything here or it will get to overwhelming which it already getting close to being. 

Who requested this? Also needs. bit of work in the details.

sstrubberg:
This was a me request. We recently just migrated the Design Career Playbook to the Gatsby Theme and I felt it should be elevated as a "Community" resource, cause, well.. it is!

We actually discussed the content of the Design Career Playbook moving into w3/design when we designed it about a year ago. Because of the "high-level" nature of the w3/design site, we felt it was a bad fit to dump the DCP into w3/design. The DCP is much more granular in content compared to w3/design.

You bring up some good points around the Switcher getting bloated. What is the barrier of entry to get something in the Switcher? In my mind, if you are a Gatsby-theme site and you were built for anything connected to Phil, the site should be listed here. Thoughts?

Also, re: the Icon - the REiD is currently internal. There is no public version (yet). Should we close this PR and wait until REiD goes public before updating the switcher?

mjabbink:
Let’s wait until REiD goes live. Just wanted this in the queue. It’s going live next week isn’t it?

Let’s not add DCP to the switcher. I really think that site and critical site need to be in w3 and I will sketch out what this looks like in the left nav. Makes total sense to be there if we’re building a design community site with w3. Can have high level stuff and tools mixed in one experience and in left nav.

jnm2377:
> You bring up some good points around the Switcher getting bloated. What is the barrier of entry to get something in the Switcher? In my mind, if you are a Gatsby-theme site and you were built for anything connected to Phil, the site should be listed here. Thoughts?

Good question, Scott. This is definitely something we should maybe evaluate and set guidelines for. I don't think that if you build with the gatsby theme, it should automatically be in the switcher, because we have SO many people who have gatsby theme sites. But idk the criteria either. I definitely think something like REID would be good to have under community.

vpicone (commented):
Should we close this and wait for the REID launch?

mjabbink:
So if this is updated in GT can we just update IDL and Carbon sites and it will have this new switcher content? If yes we need to do so since both sites are live. REiD is actually now open to the public

mjabbink:
I don’t see REiD in the switcher for Carbon or IDL.

PR #1059, opened 2020-11-12
Update ArticleCard.mdx

sstrubberg:
Working on an [issue](https://github.ibm.com/Design/racial-equity-in-design/issues/113) over in REiD around ArticleCards. The docs don't currently reflected what you can actually do with the ArticleCard.

### DIFF
--- a/packages/example/src/pages/components/ArticleCard.mdx
+++ b/packages/example/src/pages/components/ArticleCard.mdx
@@ -169,7 +169,7 @@ The `<ArticleCard>` component should generally be used inside of a `<Row>` and `
 
 | property   | propType | required | default  | description                                                                                         |
 | ---------- | -------- | -------- | -------- | --------------------------------------------------------------------------------------------------- |
-| children   | node     |          |          | Use 32x32 image as child, will display in bottom left of card                                       |
+| children   | node     |          |          | Set an image                                     |
 | href       | string   |          |          | Set url for card                                                                                    |
 | title      | string   |          |          | Card title                                                                                          |
 | subTitle   | string   |          |          | Card smaller sub title                                                                              |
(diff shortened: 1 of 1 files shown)


jnm2377:
is this just a change in the docs or are you still working on some refactoring? Wasn't sure from the issue if you were gonna make any changes to the component or not.

vpicone:
These get real wild if you put in a rectangular image or an image that's much bigger (like 128x128). Did we fix that?

sstrubberg:
lol, this is just an update to the docs to reflect what the ImageCard actually does. I had to clear up some assumptions on what ArticleCard could actually do out of the box.

jnm2377:
Hmm... I guess I'm not really understanding how `set an image` gives a different impression than `Use 32x32 image as child, will display in bottom left of card`. Unless we mean, you can now set images of any dimensions? When I read that, I still get the same impression, just with a lot less specific information (size/location where it'll appear) @sstrubberg

sstrubberg:
> Hmm... I guess I'm not really understanding how `set an image` gives a different impression than `Use 32x32 image as child, will display in bottom left of card`. Unless we mean, you can now set images of any dimensions? When I read that, I still get the same impression, just with a lot less specific information (size/location where it'll appear) @sstrubberg

Because it doesn't actually take a 32x32 image (an icon) to display in the lower left corner. It takes an image (usually 16x9) and puts it above the resource card.

I thought this was slam dunk docs change. Maybe I'm missing something. Happy to talk through it next time we get together.

jnm2377:
ooohhh, I see. I guess I just didn't understand the issue to begin with lol 👍

jnm2377 (approved):

vpicone:
@sstrubberg @jnm2377 omg I've been think this was for the resource card icon this whole time 🤦

PR #1063, opened 2020-11-18
fix(page-tabs): nav aria unique labels

andreancardona:
Closes issue found on carbon website - from theme: 

- https://github.com/carbon-design-system/carbon-website/issues/1554
-  https://github.com/carbon-design-system/carbon-website/issues/1716

**Changed**
Added. unique aria labels for tabs

### DIFF
--- a/packages/gatsby-theme-carbon/src/components/PageTabs/PageTabs.js
+++ b/packages/gatsby-theme-carbon/src/components/PageTabs/PageTabs.js
@@ -14,7 +14,7 @@ import {
 
 export default class PageTabs extends React.Component {
   render() {
-    const { tabs, slug } = this.props;
+    const { title, tabs, slug } = this.props;
     const currentTab = slug.split('/').filter(Boolean).slice(-1)[0];
 
     const pageTabs = tabs.map((tab) => {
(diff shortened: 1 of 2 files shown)


jnm2377 (approved):

vpicone:
I'm not sure the `currentTab` makes for a good label. It's not like an ID or key where we just need it to be unique. For instance [here](https://gatsby-theme-carbon-git-1716-avt1-tabs-multiple-nav-roles.carbon-design-system.vercel.app/guides/navigation/sidebar) the navigation is named `sidebar`. If you change the tab (to tabs) then the label changes to just `tabs`.

I wonder if using parent would be more appropriate so the label stays the same for that nav element between tabs (probably Navigation)

andreancardona:
> I'm not sure the `currentTab` makes for a good label. It's not like an ID or key where we just need it to be unique. For instance [here](https://gatsby-theme-carbon-git-1716-avt1-tabs-multiple-nav-roles.carbon-design-system.vercel.app/guides/navigation/sidebar) the navigation is named `sidebar`. If you change the tab (to tabs) then the label changes to just `tabs`.
> 
> I wonder if using parent would be more appropriate so the label stays the same for that nav element between tabs (probably Navigation)

okay! so changing this to `aria-label="navigation"` am I understanding that corrrectly? @vpicone

vpicone:
@andreancardona sorry it's kinda confusing/meta with these pages haha. 

As a different example, the label here would be color: https://www.carbondesignsystem.com/guidelines/color/overview/

(Not overview, usage and code)

andreancardona:
> @andreancardona sorry it's kinda confusing/meta with these pages haha.
> 
> As a different example, the label here would be color: https://www.carbondesignsystem.com/guidelines/color/overview/
> 
> (Not overview, usage and code)

ahh makes a lot more sense just pushed. up a fix @vpicone

vpicone (approved):

PR #1065, opened 2020-11-18
fix: fail gracefully without page titles

barlock:
Pages without a `title` or aren't in the `nav-items` in their front matter will fail with a very unhelpful error. Not being able to find `path` of undefined here:

```js
navigationList[navIndex].path
```

Instead, this change attempts to fail gracefully by not rendering "critical" pieces of the view. In this case, the NextPrevious (if it's not calculable), or the page header when there is nothing to header.

#### Changelog

**Changed**

- NextPrevious component gracefully handles items not in the nav
- Default layout doesn't reader a header for pages without a title


barlock:
Hmm, I'm getting this same build error on `main` is this a known issue with a fix?

packages/gatsby-theme-carbon/src/templates/Default.js
```diff
@@ -60,12 +60,14 @@ const Default = ({ pageContext, children, location, Title }) => {
...
       titleType={titleType}>
-      <PageHeader
-        title={Title ? <Title /> : title}
-        label="label"
-        tabs={tabs}
-        theme={theme}
-      />
+      {title && (
+        <PageHeader
+          title={() => <Title />}
```
barlock, line 65: Other than the fix from @vpicone, TIL that you cant just render a `<Title />` as a prop, or maybe this just expects to be a render prop? 

Either way, that is what caused that random error.

packages/gatsby-theme-carbon/src/components/NextPrevious/NextPrevious.js
```diff
@@ -185,13 +189,16 @@ const NextPreviousContainer = (props) => {
     return null;
   }
 
-  return (
+  const nextItem = getNextItem();
+  const previousItem = getPreviousItem();
+
+  return nextItem && previousItem ? (
```
vpicone, line 195: @barlock what about the last page?

packages/gatsby-theme-carbon/src/templates/Default.js
vpicone: @barlock Not rendering a page title here isn't desirable as it serves as the `h1` for the page. It's written this way to support the default case (string) as well as a [custom `Title` component](https://gatsby-theme-carbon.now.sh/guides/MDX#custom-title)

barlock: Ah ha, I wasn't aware of the custom title case.

I'm looking to, if there is no string, and no custom title component, don't render anything. 

I'll update to match that logic.

vpicone: @barlock I thought maybe the plan was to somehow infer the title from the file name? Leaving out the h1 all together is bad news a11y wise. If this is to avoid the ugly error maybe we can improve the error message rather than failing silently with a11y issues?

barlock: > I thought maybe the plan was to somehow infer the title from the file name? 

I tried this, what's interesting is my attempt at extending `onCreatePage` allowed a `develop` run to build, so the webserver would eventually start, but the rendered page still had an empty `title` (so the page header would render, but be empty). So for some reason this got passed the `NextPrevious` error, but didn't render even ugly.

> Leaving out the h1 all together is bad news a11y wise. 

Agreed, we don't want bad a11y. For pages that don't have a `title` frontmatter, but do have in their markdown a `H1`, we still have an accessible page. 

☝️ Checking this case is possible of course, but it gets complicated and would likely require a custom remark plugin (or something to check the AST). "Trust the developer" on this one seemed like an ok compromise to me.

vpicone: @barlock that's a fair point, although I don't think we have any styles for `h1` on the page, leaving that space empty feels like a mistake.

barlock: They are indeed there, header is a bit big for my taste, but I think that's just carbon:

image

---

I think I found the bug in my `gatsby-node.js` that will allow me to actually inject it. So, looks like we can calculate it. So, we have options anyway.

I can think of (assumes no title is provided):
1. Default to the base name of the file without the extension (as previously discussed, tis a bit ugly, but works, would end up with two h1s if an h1 is provided)
2. Parse the markdown, search for an H1, inject that, remove from content. 
3. Parse markdown, inject h1 IFF it's the first element, remove the header from content, fail otherwise
4. Parse markdown, search for an H1, if it can't find one, fail, but don't inject it, components don't assume a `title`, (basically don't assume a title, but check for good a11y)
5. Fail in `Default.js` with a better error message if `title === null`

---

My preference would be, in order: 4, 5, 1, 3, 2

barlock: _hmmm_, Maybe I'm coming back around to 1

vpicone: @barlock yeah I agree, I think one is best. Anything to do with parsing feels a bit fragile

packages/example/src/pages/contributions.mdx
```diff
@@ -1,16 +1,19 @@
 ---
-title: Contributions
 description: Instructions for developing and contributing to the theme
 ---
 
+<!-- Page title left out so it can be generated -->
+
 <PageDescription>
 
+
```
barlock, line 9: Prettier added these spaces.. Not really sure why now though.

packages/gatsby-theme-carbon/gatsby-node.js
```diff
@@ -1,6 +1,7 @@
 const path = require('path');
 const fs = require('fs');
 const mkdirp = require('mkdirp');
+const startCase = require('lodash.startcase');
```
barlock, line 4: Possibly controversial? `gatsby-plugin-mdx` includes the whole package already. I don't think this will be bundled to the browser.

vpicone: Yeah, shouldn't be a problem

barlock:
Anything holding this back?

packages/gatsby-theme-carbon/src/components/NextPrevious/NextPrevious.js
```diff
@@ -150,6 +150,10 @@ const NextPreviousContainer = (props) => {
   };
 
   const getNextItem = () => {
+    if (!navigationList[navIndex]) {
```
jnm2377, line 153: Is this still needed if you're getting the title from the filename if one doesn't exist?

packages/gatsby-theme-carbon/src/components/NextPrevious/NextPrevious.js
```diff
@@ -185,13 +189,16 @@ const NextPreviousContainer = (props) => {
...
+  return nextItem && previousItem ? (
     <NextPrevious
       isHomepage={isHomepage}
-      previousItem={getPreviousItem()}
-      nextItem={getNextItem()}
+      previousItem={previousItem}
+      nextItem={nextItem}
     />
-  );
+  ) : null;
```
jnm2377, line 201: same as above

packages/gatsby-theme-carbon/src/components/NextPrevious/NextPrevious.js
barlock: No, I'm surprised actually. This was a semi related challenge where the page wouldn't render if it wasn't in the nav list even if it had a title. (I often do this to dark launch pages)

Reverse code gremilins! 🚀

jnm2377 (approved):
Sorry, it's taken so long to approve this. I was out for a bit during the holidays. Thanks for working with us on this! LGTM 🎉

barlock:
@jnm2377 What else do we need to do for this to be merged?

jnm2377:
@vpicone can you review?

vpicone (approved):

PR #1077, opened 2020-12-16
style: removed the padding -closes #1074

### LINKED ISSUE #1074: Content & code snippet in list should go to the right margin
————

sstrubberg:
Closes #1074

### DIFF
--- a/packages/example/src/pages/contributions.mdx
+++ b/packages/example/src/pages/contributions.mdx
@@ -6,13 +6,11 @@ description: Instructions for developing and contributing to the theme
 
 <PageDescription>
 
-
 Use this guide to contribute to the theme. We’ll show you how to get the
 development environment set up as quickly as possible so you can start
 contributing.
 
 </PageDescription>
-
 
 <AnchorLinks>
   <AnchorLink>Project setup</AnchorLink>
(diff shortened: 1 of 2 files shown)


sstrubberg:
@mjabbink is this what you're wanting? Your spec wasn't clear on what it needed to look like for md and lg screens.

mjabbink:
Looks right to me. Those images (text and snippet) in issue were meant to illustrate that they needed to go to the right margin versus being cut short. Looks like you resolved that and unfortunately I forgot to paste in the link where I screen grabbed those.

vpicone (commented):
I think we only want to make this tweak on mobile right?

mjabbink:
Hmm. In desktop something looks off as well. The right side of list items should not exceed the typical 8 column width for the main body spans across. It should also not get padding on the right side. So only left side is getting indented as the list needs to with bullets, numbers letters, etc but the content always spans to the same right edge of the margin same as NON list body copy. Hope this makes sense.

I the sample I pointed out in the issue the content was getting extra padding on the right so it was spanning to the right edge.

In desktop (images below) I can’t tell if the snippet containers is too short or if the text is what is going too long?
https://gatsby-theme-carbon-git-issue1074.carbon-design-system.vercel.app/getting-started

Screen Shot 2020-12-16 at 3 15 22 PM


———

mjabbink:
Just double checked. The above has the text going beyond the 8 columns and the image above that has the snippet going beyond the 8 columns. This means something wonky is going on.

jnm2377 (approved):

jnm2377:
Hmm.. maybe we should get specs for the list if there are more issues with it.. I thought this issue was just about padding but seems like there are more things wrong with list alignment/width/etc.

sstrubberg:
Left a comment on the original issue regarding the specs.

sstrubberg:
Went ahead and did Vince's suggestion. I need feedback plz.

sstrubberg:
turned out the code snippets were in the lists, which is not a pattern we normally do. I just pulled them out and manually made the ordered list.

PR #1080, opened 2020-12-21
feat(code): add configurable collapsing

barlock:
Adds a parameter `overflow` to the `Code` component to make the overflow point configurable.

#### Changelog

**New**

- Adds a parameter `overflow` to the `Code` component to make the overflow point configurable.


packages/gatsby-theme-carbon/src/components/Code/Code.js
```diff
@@ -76,14 +81,14 @@ const Code = ({ children, className: classNameProp = '', path, src }) => {
...
+      {isOverFlowEnabled && shouldOverflow && (
         <button
           className={cx(styles.showMoreButton, {
             [styles.dark]: interiorTheme === 'dark',
           })}
-          onClick={() => setShouldShowMore(!shouldShowMore)}
+          onClick={() => setOverflowExpanded(!isOverflowExpanded)}
           type="button">
-          {shouldShowMore ? (
+          {isOverflowExpanded ? (
```
barlock, line 91: I refactored these booleans because the code was confusing to me.

Before to me it read as: `If I should show more, render the component that says "Show Less"`

I made it a little more clear what each bool represented.

packages/example/src/pages/components/code-blocks.mdx
```diff
@@ -56,3 +80,4 @@ This code snippet provides both a `path` and a `src`.
 | language | string   | [Available languages.](https://github.com/FormidableLabs/prism-react-renderer/blob/master/src/vendor/prism/includeLangs.js) |
 | src      | string   | The full url of a relevant link (source)                                                                                    |
 | path     | string   | A string indicating the filename or path. Due to markdown limitations this can only be a single word                        |
+| overflow | number   | A number indicating how many lines to display before overflowing. Use `0` to disable overflowing completely.                |
```
barlock, line 83: My original purpose with this change was to be able to disable overflowing (it doesn't make sense in all contexts), making it configurable seemed good to me. That would enable an author to say "Well, if I have 10 lines, maybe I shouldn't overflow just one line".

packages/gatsby-theme-carbon/src/components/Code/Code.js
```diff
@@ -14,9 +14,18 @@ import Sidebar from './Sidebar';
...
-  const [shouldShowMore, setShouldShowMore] = useState(false);
+const Code = ({
+  children,
+  className: classNameProp = '',
+  overflow = 9,
+  path,
+  src,
+}) => {
+  const isOverFlowEnabled = overflow > 0;
+  const shouldOverflow = children.split('\n').length > overflow;
```
barlock, line 25: I moved this out of `getLines` to fix this warning:
image

sstrubberg (changes requested):
Loving the intention behind this PR. Well played. I noticed some styling needs fixing when I applied the `overflow=0`. 

iconMouseOver

Also, not sure if there is another way to communicate how to use the overflow. It wasn't very intuitive (i had to just open the MDX and see for myself).

barlock:
I could only reproduce that gif by creating a markdown block with no code in it. I went ahead and added a `min-height` to fix that bug. I'm thinking there's likely something else wrong though? 

Could you provide the code you used to make that gif if this change doesn't fix whatever your seeing?

barlock:
> Also, not sure if there is another way to communicate how to use the overflow.

English isn't my strongest language (JS tends to be), any suggestions to the docs? I took another pass at making them more explicit.

barlock:
@sstrubberg bump! I think I fixed the styling issues but I'd love to know if things are looking better.

sstrubberg:
Docs are much better. Thx!

I'm still having trouble with the collapsing. I also discovered a new bug in green where there is sandwich'ing of icons going on.

image

barlock:
Got spot on the bug! I'll get a fix in for that. For the overlow not rendering right... do you need to clean your cache?

I still can't reproduce and the app deployed by vercel doesnt have that issue: https://gatsby-theme-carbon-git-fork-barlock-optional-collapse.carbon-design-system.vercel.app/components/code-blocks

Obviously I don't want you to just trust me blindly, but I'm not sure what's going on there.

barlock:
I found an interesting thing where MDX params are always parsed as strings 🤷‍♂️ Perhaps that was doing something screwy on your environment? I typecast it as I needed to for another bool check. 

When you set it to `1` I went with "inlining" the icon on that one line. I don't know why anyone would want to set it to 1, but 🤷‍♂️ that seemed like the best style to me. 

Looks like this:
image

Naming, for scss modules should I use BEM? It looks weird in JS land, wasn't sure what ya'lls standards were.

sstrubberg:
I like what you did to fix the overlapping icons when the overflow is set to `1`. Can you do the same thing for when it's set to `2`? It looks like there is still a bit of overlap there.

image


Regarding my local environment and parsing of things. After a little investigating, once I disabled Prettier, I wasn't having the issue any more. So now I'm torn... I don't want to have to disable prettier to take advantage of this feature. I'm also open to this being an issue with only my machine and my prettier config. I think we need an outside perspective @vpicone, care to weigh in? To catch you up.. 1) Do you have prettier installed on your VSCode? If so, then 2) Try and change the overflow from 5 to 0 in Michael's example. On my machine, the code block content will just straight up disappear when I prettier enabled. When I have it disabled, it works just fine.

barlock:
That's weird, I also use prettier and it should be updating on save. How is it changing the source to no longer pick up the param? Can you give a "test case"? Like a known failing source file (presumably one updated by prettier) that should work? 

Getting something that's prettier compatible is important.

vpicone:
Woah! Thanks for the PR @barlock this is really neat. In my opinion, specifying the amount to overflow (rather than the number of visible lines) feels a bit backwards.

Intuitively, it's easier for me to figure out how many lines I _want_ to show rather than how many I _don't_. I think this becomes more true if the code snippet is updated and the number of lines is changed.

barlock:
That’s what it does. I used the variable name y’all had I think? What name
do you think would be better for a public api? overflowAfter?

vpicone:
@barlock I was thinking instead someone would specify the number of _visible_ lines rather than counting the number overflowed.

barlock:
> I was thinking instead someone would specify the number of visible lines rather than counting the number overflowed.

Yes, I agree. The code _should_ be doing that. If it doesn't I would call it broken. I've changed the variable from `overflow` to `overflowAfter` to imply: "I would like my code to overflow after displaying X lines".

Does this work for you? Do you have any other suggestions for variable names?

If my explanation doesn't make sense then can you maybe give me a more concrete example or a proposed API that I can implement against?

vpicone:
@barlock right on, that makes sense. I like overflowAfter better.

packages/gatsby-theme-carbon/src/components/Code/Code.js
```diff
@@ -14,9 +14,19 @@ import Sidebar from './Sidebar';
...
+  className: classNameProp = '',
+  overflow: overflowProp = 9,
+  path,
+  src,
+}) => {
+  const overflow = Number(overflowProp); // MDX parses this as a string
+  const isOverFlowEnabled = overflow > 0;
+  const shouldOverflow = children.split('\n').length > overflow;
+  const [isOverflowExpanded, setOverflowExpanded] = useState(
+    !isOverFlowEnabled
```
vpicone, line 28: Why does this default to `!isOverFlowEnabled`?

packages/gatsby-theme-carbon/src/components/Code/Code.js
```diff
@@ -33,15 +43,11 @@ const Code = ({ children, className: classNameProp = '', path, src }) => {
   const getLines = (lines) => {
     const withoutTrailingEmpty = removeTrailingEmptyLine(lines);
 
-    if (withoutTrailingEmpty.length > 9) {
-      setHasMoreThanNineLines(true);
-    }
-
-    if (shouldShowMore) {
+    if (isOverflowExpanded) {
```
vpicone, line 46: Need to normalize the capitalization between variagbles `OverFlow` vs `Overflow`

vpicone (changes requested):

packages/gatsby-theme-carbon/src/components/Code/Code.js
barlock: The thought was "If overflow is disabled, then, by default, 'expand the overflow'" 

and 

```
isOverflowEnabled ? true : false
```

simplifies to `!isOverFlowEnabled`

_However_ after forgetting about this change 😅 and myself not being able to figure that out in 30 seconds, I rewrote this to move that check into the `getLines` function instead which is probably more clear.

barlock:
Apologies for the delay in getting this up. Much distractions popped up. Should be ready again.

packages/gatsby-theme-carbon/src/components/Code/Code.js
barlock: 🤦 Nice catch

vpicone:
Hey @barlock, sorry this one's spiraled on too long. I'm not sure this feature is worth the additional complexity at the moment. There's a few [rules we have](https://github.com/carbon-design-system/gatsby-theme-carbon/blob/main/docs/development-principles.md#what-should-be-a-feature) for what gets to be a feature. If you want to implement it in your own site via shadowing so we can demonstrate value, I think that would be the best first step.

barlock:
Been using it in production since the PR was made almost 6 months ago. I just turn off overflow everywhere which is right for my use-case.

I'm a lot disappointed in your "should it be a feature" rationale. My gut reaction is to stop making PRs as they appear to not ever be accepted. Plus given the burden of shadowing on breaking changes it makes it that much harder to maintain a fork, it becomes significantly harder to even use this theme. 

Help me help you, how can we both avoid this much wasted time? I've [opened issues](https://github.com/carbon-design-system/gatsby-theme-carbon/issues/1084) that haven't been addressed by a maintainer in months so I can't even discuss the code before making the PR. It all feels disheartening.

PR #1091, opened 2021-01-20
feat: last modified date on pages

sstrubberg:
Users now have the option of enabling or disabling a feature to display the last time a page was modified and committed to git on Inner pages, giving users clarity on when the last time a given page was updated.

image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,4 @@
 # Logs
-logs
-*.log
 npm-debug.log*
 yarn-debug.log*
 yarn-error.log*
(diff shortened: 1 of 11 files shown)


jeanservaas:
@sstrubberg 

This is great, I was just wondering, in terms of the design, if we could stack these two pieces of information at the bottom of the page when people do enable the last modified feature, just to tighten up the design of it a bit

image

jnm2377 (commented):
Nice! This is really cool 🎉 is it still WIP or are you done?

PR #1094, opened 2021-01-27
feat: date in frontmatter resolves last modifed date on page

sstrubberg:
Through the page frontmatter, Gatsby-Theme users can now enter a date that the page was last modified. This date will then be displayed at the bottom left corner of the page. If no date is specified, nothing will be generated at the bottom of the page.

image

---

I also cleaned up the publishing docs while I was in there.


jnm2377 (approved):

vpicone (commented):
This is so clean and awesome, killer work. One small idea then one bigger idea:

---

Is there a reason we're calling the component _modified_ date? 

&&

I think there's a good opportunity to introduce a _little_ bit of opinion here so that the formatting doesn't get sloppy/out of sync between posts (and sites). If we create a new date object using their string `const lastUpdated = new Date(frontmatterDateString)` then we can use [toLocaleDateString](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleDateString) to give them the appropriate formatting for their browsers locale:

`lastUpdated.toLocaleDateString()`

Maybe get with Jan to figure out which [`dateStyle`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat/DateTimeFormat#parameters) option would be best.

sstrubberg:
@vpicone I got with Jan and settled on the date formatting. All should be good there. I was curious on your thoughts on what the component should be called instead of `<ModifiedDate>`. When I call the function simply `Date` it conflicts with the existing `Date` function that I'm calling process the localeDateToString.

vpicone:
@sstrubberg Ahhh that makes sense, I was thinking you meant like a modified version of `Date`. Maybe LastModifiedDate? Either way I agree `Date` is bad, don't want to use existing keywords.

packages/gatsby-theme-carbon/src/components/ModifiedDate/ModifiedDate.js
```diff
@@ -0,0 +1,25 @@
...
+    month: 'long',
+  };
+
+  const lastModified = new Date(date);
+
+  return date ? (
+    <Row className={styles.row}>
+      <Column>
+        <div className={styles.text}>
+          Page last updated: {lastModified.toLocaleDateString('en-GB', options)}
```
vpicone, line 18:
```suggestion
          Page last updated: {lastModified.toLocaleDateString(undefined, options)}
```

We don't want to force the English locale. By using `undefined` (no quotes), it will defer to the browser's setting.
Screen Shot 2021-02-08 at 10 37 52 AM

vpicone (changes requested):

vpicone (approved):

PR #1096, opened 2021-01-29
fix: zindex for search bar on headernav style

sstrubberg:
More z-index issues with the header. Theres probably a more elegant way to do this. I feel like i'm playing wack-a-mole where my z-index property is the hammer and the onClick is the mole.

Currently, the nav items in the header nav aren't able to be clicked on because of their z-index. This fixes that by altering the z-index.

### DIFF
--- a/packages/example/gatsby-config.js
+++ b/packages/example/gatsby-config.js
@@ -27,6 +27,7 @@ module.exports = {
             'https://github.com/carbon-design-system/gatsby-theme-carbon',
           subDirectory: '/packages/example',
         },
+        navigationStyle: 'header',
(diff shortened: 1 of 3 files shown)


jnm2377 (commented):
Yeah, the z-index changes are getting a little confusing, but this seems to be working. Could you maybe add some comments explaining the zindex changes for future maintainers? Something like `//when search is open, //when search is closed, etc.`?

jnm2377 (approved):
🎉 thanks for adding the comments, it's really helpful!

vpicone (approved):

PR #1097, opened 2021-02-03
Update ResourceCard.mdx

mzuliani-ibm:
#### Changelog

Add `none` as a valid option for actionIcon attribute.


vpicone:
@jeanservaas is it cool for resource cards to not have an icon?

PR #1106, opened 2021-03-17
fix: update latest carbon and add style override for sidenav overlay bug

jnm2377:
rel: https://github.com/carbon-design-system/carbon-website/pull/2187

cc: @tw15egan


packages/gatsby-theme-carbon/package.json
```diff
@@ -33,8 +33,8 @@
     "@reach/router": "^1.2.1",
     "@vimeo/player": "^2.9.1",
     "beautiful-react-hooks": "^0.23.1",
-    "carbon-components": "^10.23.2",
-    "carbon-components-react": "^7.23.2",
+    "carbon-components": "^10.30.0",
```
tw15egan, line 36: I'm not sure if `Dropdown` is used in the theme, but I know `10.30.0` has that chevron bug, so we might want to just bump this to `10.31.0`

tw15egan (approved):

vpicone (approved):

PR #1119, opened 2021-04-16
chore: update carbon, remove switcher tooltip

vpicone:


mjabbink:
@vpicone Was this update deployed? It’s still visible in GT demo site and across the sites

mjabbink:
OK, Alison caught me up on this. All good with exception of the name. In the preview, the tooltip says “App switcher”

I reckon just “Switcher” or “Site switcher” could work. @jeanservaas

janchild:
@vpicone @alisonjoseph I'm not sure what this original PR was supposed to be. It says remove it, but I checked and the tooltip is still there although now it says `Switch` which would not have been my choice.  

If we need a tooltip, I would use `Site switcher`, as @mjabbink suggested above. 

Also we should use the tooltip with the arrow at the end. to avoid the tooltip edge disappearing under the scrollbar. 

image

mjabbink:
I think “Site switcher” will do the trick.

vpicone:
> @vpicone @alisonjoseph I'm not sure what this original PR was supposed to be. It says remove it, but I checked and the tooltip is still there although now it says `Switch` which would not have been my choice.  
> 
> 
> 
> If we need a tooltip, I would use `Site switcher`, as @mjabbink suggested above. Or, `Switch sites` so we are speaking directly to the user. This second option might be preferable. Any thoughts?
> 
> 
> 
> Also we should use the tooltip with the arrow at the end to avoid the tooltip edge disappearing under the scrollbar. Looks like this one has the arrow in the center. 
> 
> 
> 
> image

That change in visibility came from upstream in Carbon UI shell, icon buttons now have a mandatory tootip for accessibility. 

The change to the tooltip contents is published in the theme, we just need to update the theme in the Carbon website.

PR #1130, opened 2021-06-05
Configuration for enabling or disabling Switcher

sid-r-singh:
Please provide configuration for enabling or disabling Switcher. Otherwise just like how you provided 
shadowed versions of Navlinks, Header & Footer if you can provide shadowed version of Switcher
it would be great, so that the links can be modifed.


sstrubberg (changes requested):
I'm a bit confused by this PR. When i first read the title, it looked like you were going to add the logic needed for enable or disable the switcher, but then it turned out to be guidance in the form of MDX, but I wasn't even clear on what kind of guidance you were suggesting for people. Can you elaborate more on this please :)

vpicone:
This is a not a legitimate pull request, it looks like you're just requesting documentation by using dummy content? In the future, just open the issue and leave it at that.

PR #1151, opened 2021-07-26
DO NOT MERGE - feat(square-card): add new square card component 

### LINKED ISSUE #1061: Enhancement[card]: add 1:1 card to component library
## Enhancement: the 1:1 Card

@DianaStanciulescu is currently updating the Gatsby kit and there are several high priority component enhancements that we would like to add to our Gatsby component library. 

The specs are below, so this is solely a dev work stream that's needed.

### Justification

Although the 1:1 card with no image is being used widely in Gatsby sites and in product, it does not exist as a component. As a result teams need to modify the 1:1 image component to achieve this card, but more importantly, they don't understand type hierarchy in this card. Offering type hierarchy variants alone with drastically decrease support questions for Gatsby theme.

Some use cases:

https://www.ibm.com/able/
https://www.ibm.com/able/toolkit
https://www.ibm.com/design/thinking/page/courses/Practitioner
https://www.ibm.com/design/thinking/page/toolkit
https://www.ibm.com/design/research/guiding-principles/principles-overview
https://www.ibm.com/standards/web/carbon-for-ibm-dotcom/

### Design specs and variants

Ideally this card configuration can work with an H2 title or an H3 title.

image

## The is an enhancement of the image component so grid behavior is exactly the same as it is for that component.

image

### Sketch file attached

[1:1 Card Specs.zip](https://github.com/carbon-design-system/gatsby-theme-carbon/files/5548829/1.1.Card.Specs.zip)

_This is a brand prioritized issue so that Diana's kit matches what is available in the code_

aledavila:
Closes #1061 

DO NOT MERGE - need to fix mobile border issues with cards 

#### Changelog

**New**

- Added Square Card component

- Considered and enhancement from image but should be considered its own component. 
- To test make check Square Card page


aledavila:
@jeanservaas  @vpicone @sstrubberg Its ready for review now :)

sstrubberg:
image

Might be a good idea to drop to a small type token at large screens.

aledavila:
I think at larger screens its supposed to be large type. But yea theres a certain range where it overlaps @jeanservaas. How do you think this should be handled. There a small window where it looks like this right before and wraps to 2 columns.

jeanservaas:
@sstrubberg I think what you guys have done with the adaptive approach is good for the two square cards that have sentences. Wondering if for this "short title" one, if it would work to have it stay at $productive-heading-04 at all breakpoints instead of jumping down to $productive-heading-03.

image

Also can we just shorten the small bottom line to say (Optional text)?

I don't really want people writing a lot here, it's more appropriate for say, a date or a category name than a sentence.

image

jeanservaas (commented):
See comment below

aledavila:
@jeanservaas It appears the the headings are 32px. I’m using productive-04. The computed font-size for the cards says it’s 28px. I think the example headings is what is incorrect are those supposed to be 28px I can submit a change for that. 

Also to answer your question one thing we can do is add a prop that if a user adds that prop it’ll be productive-04 on all screen sizes. Like a “shortTitle” prop for example or instead do default productive-04 with a prop “sentenceTitle” that would make it productive-03. That would be the easiest solution. Otherwise you’d have to make a function to consider text length but the cleanest would be the prop. What are your thoughts on that ?

sstrubberg (approved):

alisonjoseph:
@aledavila @jeanservaas can we get this finished up next sprint? seems like its close?

jeanservaas:
Hey Ale, we either need to drop these Optional type moments at smaller sizes, break the line sooner (i.e there'd be about 24px of space between the icon and the end of the type), or cut the extra type, we can't allow this crunch up to happen at small type sizes.

image


Also, for this adaptive scenario, the space after the type is fine, because these are two different type sizes

image

but beyond the large breakpoint, when the type sizes become the same, there would be no extra space between the heading and the body copy

image

**So it would look like this if we go down a type size**

image

All else looks good. @aledavila can you ping me when you get this done and I'll give you a review really fast. Sorry.

jeanservaas (approved):
Looks good!

packages/gatsby-theme-carbon/src/components/SquareCard/square-card.scss
```diff
@@ -0,0 +1,185 @@
+// Square Card Group
+.square-card-group .#{$prefix}--col-lg-4 {
+  margin-bottom: 0;
+}
+.square-card-group .#{$prefix}--square-card {
```
vpicone, line 5: We want to use CSS modules for new components. We have some [contribution notes](https://gatsby.carbondesignsystem.com/contributions#sass-and-css-modules) on CSS modules, the other components can help for reference too.

packages/gatsby-theme-carbon/src/components/SquareCard/square-card.scss
```diff
@@ -0,0 +1,185 @@
+// Square Card Group
+.square-card-group .#{$prefix}--col-lg-4 {
```
vpicone, line 2:
```suggestion
.square-card-group > *
```

Better to be less specific in case the number of columns changes. We're really just trying to target all the direct children, we don't super care the number of columns.

packages/gatsby-theme-carbon/src/components/SquareCard/square-card.scss
```diff
@@ -0,0 +1,185 @@
...
+  margin-bottom: 0;
+}
+.square-card-group .#{$prefix}--square-card {
+  border-top: 1px solid #e0e0e0;
+
+  @include carbon--breakpoint('md') {
+    border-left: 1px solid #e0e0e0;
+  }
+}
+.square-card-group .#{$prefix}--col-lg-4 .#{$prefix}--square-card {
```
vpicone, line 12: Do we need all this specificity? Could it just be `.#{$prefix}--square-card` (or `.squareCard` with CSS modules)?

packages/gatsby-theme-carbon/src/components/SquareCard/square-card.scss
```diff
@@ -0,0 +1,185 @@
...
+.square-card-group
+  .#{$prefix}--col-lg-4:nth-child(3n + 1)
+  .#{$prefix}--square-card {
+  @include carbon--breakpoint('lg') {
+    border-left: 1px solid transparent;
+  }
+}
+
+// Square Card Component
+.#{$prefix}--square-card {
```
vpicone, line 52:
```suggestion
.squareCard {
```

packages/gatsby-theme-carbon/src/components/SquareCard/SquareCard.js
```diff
@@ -0,0 +1,180 @@
...
+          {actionIcon === 'disabled' || disabled === true ? (
+            <Error20 aria-label="disabled" />
+          ) : null}
+        </div>
+      </>
+    );
+
+    let cardContainer;
+    if (disabled === true) {
+      cardContainer = <div className={carbonTileclassNames}>{cardContent}</div>;
```
vpicone, line 96: Should this just be the Carbon Tile component?

vpicone (changes requested):

packages/gatsby-theme-carbon/src/components/SquareCard/square-card.scss
aledavila: Oh ok. I was referencing how the other cards were built

vpicone: Ah that makes sense, those were old components we never migrated over. Let me know if you run into any issues or wanna pair on some of it

vpicone (approved):
Looks great! Awesome job. Thanks for making those changes.

PR #1153, opened 2021-07-29
fix(page-tabs): ensure scrollbar is dark

theiliad:
if you're on light theme, even though page tabs render as dark, the scrollbar still renders in light

this fixes that

**before**

image

**after**

image

### DIFF
--- a/packages/gatsby-theme-carbon/src/components/PageTabs/PageTabs.module.scss
+++ b/packages/gatsby-theme-carbon/src/components/PageTabs/PageTabs.module.scss
@@ -15,6 +15,7 @@
   padding: 0;
   overflow-x: auto;
   scrollbar-width: none;
+  color-scheme: dark;
 }
 
 .list-item {


sstrubberg (approved):
What if the user chooses the white/white theme?

theiliad:
> What if the user chooses the white/white theme?

Do the page tabs ever render in white? I haven't seen that anywhere.

If yes, then we'd have to only apply this when rendering in dark

If no, then we should be good with this change

jeanservaas:
@sstrubberg do you mean like a situation like this, when using the gatsby white theme and potentially having a white banner?

image

theiliad:
> @sstrubberg do you mean like a situation like this, when using the gatsby white theme and potentially having a white banner?
> 
> image

is there an actual gatsby white theme? I seem to only find examples that match carbondesignsystem.com

jeanservaas:
@theiliad Yes:

https://w3.ibm.com/design/community

theiliad:
> @theiliad Yes:
> 
> 
> 
> https://w3.ibm.com/design/community
> 
> 
> 
> That was built before Scott actually did the theming work, but white theme should now be an option in Gatsby.

Nvm then

PR #1155, opened 2021-07-29
fix(code-module): ensure the scrollbar is dark

theiliad:
related to #1153 

seems like this component also always renders in dark, so this PR ensures that the dark scrollbar always shows up inside it

**before**

image

**after**

image


sstrubberg (approved):

theiliad:
@sstrubberg should we get someone else to review? Weird, I don't have access in the repo to request more ppl

PR #1160, opened 2021-08-19
feat(minicard): support transferring props to Gatsby Links and anchors

### LINKED ISSUE #1159: Transfer props to Gatsby Links and anchors in MiniCards
### Summary

The MiniCard component currently does not support additional props for Gatsby Links or anchors contained within it. Adding support for transferring props to MiniCard links would allow users to use more of the Gatsby Link API or anchor element attributes (e.g. https://github.com/carbon-design-system/carbon-for-ibm-dotcom-website/issues/1147)

### Available extra resources

What resources do you have to assist this effort?

I can open a PR to address this

emyarod:
Closes #1159

This PR adds support for transferring props to Gatsby Links and anchors in the MiniCard component via the `linkProps` prop

#### Changelog

**New**

- `linkProps` prop on `<MiniCard />`

**Changed**

- typo in MiniCard docs

### DIFF
--- a/packages/example/src/pages/components/MiniCard.mdx
+++ b/packages/example/src/pages/components/MiniCard.mdx
@@ -10,14 +10,12 @@ description:
 
 <PageDescription>
 
-
 The `<MiniCard>` component can be used in place of a `<ResourceCard>` if your
 content allows it. Unless it is sitting aside your main conent, it should always
 be wrapped inside of a `<CardGroup>`. This allows the correct gutter and border
 placement between a group of cards.
 
 </PageDescription>
-
 
 <InlineNotification>
   Although the mini-resource card has a similar geometry to the button
@@ -33,15 +31,12 @@ placement between a group of cards.
 
 <CardGroup>
 
-
 <MiniCard title="26 characters per MiniCard" href="/demo">
-
 
 ![Adobe Acrobat icon](/images/adobe-icon.svg)
 
 </MiniCard>
 <MiniCard title="Use the default icon" href="/demo" />
-
 
 <MiniCard
   title="Choose from other icons"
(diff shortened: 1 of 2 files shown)


sstrubberg (approved):

alisonjoseph:
@vpicone can we get this merged and published?

vpicone (changes requested):
Can we just use the already existing `rest` prop since it's being passed to the most exterior component?

emyarod:
> Can we just use the already existing `rest` prop since it's being passed to the most exterior component?

no that doesn't work because it doesn't affect links and anchors in MiniCards (#1159), unless I'm misunderstanding you

vpicone:
@emyarod the cards themselves are links, so you would just pass the additional props to the card as if you were passing them to a Link component. We want to avoid the `internalcomponentProps` pattern. We already do this for [ArticleCard](https://github.com/carbon-design-system/gatsby-theme-carbon/blob/5201116919b34cb42e4875f4a99b0904ae49fbaf/packages/gatsby-theme-carbon/src/components/ArticleCard/ArticleCard.js#L107) for example. Doing `linkProps` would deviate from the existing pattern.

emyarod:
@vpicone the MiniCards links are wrapped in a grid column div which doesn't pass additional props to the links

https://github.com/carbon-design-system/gatsby-theme-carbon/blob/4c125ca381601c52b35236c5d43c3c6afa58085e/packages/gatsby-theme-carbon/src/components/MiniCard/MiniCard.js#L70

in the MiniCards example `gutterLg` is passed into the MiniCard to modify the grid column and is not passed into the links which I think is expected. Currently there is no way to add props to the links unless I am misunderstanding you

https://github.com/carbon-design-system/gatsby-theme-carbon/blob/main/packages/example/src/pages/components/MiniCard.mdx?plain=1#L74-L78

vpicone:
@emyarod fair enough, thanks for clarifying

vpicone (approved):

emyarod:
@vpicone has this change been published in a new 1.x release?

vpicone:
@emyarod it won't be published to a 1.x release, we're not backporting issues. The improvements to build times and dev start up times are worth the migration if you haven't yet. Should be published to the 2.x branch by eod.

PR #1172, opened 2021-10-05
fix: line length for lists and headings

### LINKED ISSUE #1156: Body text column too wide with when using a “List” Copy should exceed the 8th column
The body copy exceeds the 8th column when using a list. The copy should NOT go beyond the normal width of the body copy. So either with **body copy** and **List copy**, the rights side (or length of copy) should never exceed the 8th column of the grid.

alisonjoseph:
Closes #1156 

Adds correct right padding to lists and headings

#### Test

Make sure nothing with lists or headings in markdown are broken

(screen shot from local with longer headings)
Screen Shot 2021-10-05 at 2 04 14 PM

### DIFF
--- a/packages/gatsby-theme-carbon/src/components/markdown/Markdown.module.scss
+++ b/packages/gatsby-theme-carbon/src/components/markdown/Markdown.module.scss
@@ -44,15 +44,14 @@
 .h3,
 .h4,
 .h5,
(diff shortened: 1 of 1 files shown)


vpicone (approved):

vpicone:
Nice!

PR #1183, opened 2021-10-26
feat: add motion-spec component

janhassel:
Adds a new component, `<MotionSpec />` that allows for standardized motion documentation on the Carbon website and our PALs.

It accepts a prop, `spec`, which is structured as a list of elements with a list of their respective transitions.

Example of a fictional dropdown, describing the expanding motion:

```js
const spec = [
  {
    element: 'Chevron',
    transitions: [
      {
        property: 'Rotation',
        curve: 'standard',
        mode: 'productive',
        duration: 112,
        delay: 48,
        from: '0°',
        to: '180°',
      },
    ],
  },
  {
    element: 'List',
    transitions: [
      {
        property: 'Opacity',
        curve: 'entrance',
        mode: 'productive',
        duration: 160,
        delay: 0,
        from: '0%',
        to: '100%',
      },
      {
        property: 'Height',
        curve: 'entrance',
        mode: 'productive',
        duration: 160,
        delay: 0,
        from: '0',
        to: '16rem',
      },
    ],
  },
];
```

It uses this information to render a timeline-like representation of the motion composition.

image

This timeline allows both designers and developers to quickly understand and apply or implement the described motion.

Carbon's duration tokens are supported, meaning that one can also define a duration as `moderate-01` instead of `150ms` and it will render accordingly. If one defines a number which matches a duratiobn token it will be displayed as the token automatically (writing `duration: 110` will result in `fast-02` being shown).

#### Changelog

**New**

- Added component: `MotionSpec`


sstrubberg:
@vpicone @janhassel shall we go ahead a close this PR for now?

Issue #1083, opened 2021-01-05
Enhancement[ expressive list]: add expressive list to component library


DianaStanciulescu:
The "expressive list" pattern is used a lot throughout gatsby sites. There are variations in how the rule is designed, spacing and behavior on tablet, so a component with predefined specs would help consistency. 

examples of the list in use:
- in full width, top nav, pages:
https://www.ibm.com/design/practices/
https://www.ibm.com/design/teams/
https://www.ibm.com/design/impact/

- in pages with a left nav:
https://w3.ibm.com/design/about/overview/
https://w3.ibm.com/design/leadership/overview/
https://w3.ibm.com/design/community/
https://www.ibm.com/brand/architecture
https://www.ibm.com/brand/systems/services/strategy
https://www.ibm.com/brand/systems/it-infrastructure/expression/strategy/
https://www.ibm.com/design/language/philosophy/principles/
https://www.ibm.com/design/language/layout/overview/
https://www.ibm.com/design/language/illustration/overview/ (with a white background)
https://www.ibm.com/able/toolkit/plan (with a white background and a pictogram)

specs excerpts:

<img width="1049" alt="Screenshot 2021-01-05 at 11 56 45" src="https://user-images.githubusercontent.com/51086727/103632600-35493b80-4f4d-11eb-8808-4187080c96d6.png">
<img width="1020" alt="Screenshot 2021-01-05 at 11 57 01" src="https://user-images.githubusercontent.com/51086727/103632608-367a6880-4f4d-11eb-999e-cb86247563b6.png">
<img width="832" alt="Screenshot 2021-01-05 at 11 57 14" src="https://user-images.githubusercontent.com/51086727/103632610-37ab9580-4f4d-11eb-8107-27c692e2a95d.png">

sketch file:
[Expressive list specs_ds3.zip](https://github.com/carbon-design-system/gatsby-theme-carbon/files/5769401/Expressive.list.specs_ds3.zip)

DianaStanciulescu:
FYI  @vpicone @jeanservaas @mjabbink

mjabbink:
@DianaStanciulescu Beautiful. These will be super usefu/helpful additions to the theme.

mjabbink:
FYI @alisonjoseph @sadekbazaraa

DianaStanciulescu:
I don't know anything about where these new gatsby components stand, they probably need to be prioritized on the dev side.

mjabbink:
[...]

PR #1197, opened 2021-12-09
feat(expressive-list): creates the expressive content only list

aledavila:
Ref #1083

This issue needs to be divided into separate components as styles and usage are different.
Creates expressive list for content only with background.

Screen Shot 2021-12-13 at 2 53 38 PM
Screen Shot 2021-12-13 at 2 53 41 PM

#### Changelog

**New**

- `ExpressiveList`


packages/example/src/pages/components/ExpressiveList.mdx
```diff
@@ -0,0 +1,135 @@
...
+| ---------- | -------- | -------- | -------- | ------------------------------------------------------------------------------------------------------------------ |
+| background | bool     |          | false    | Set true to add a background. Theme aware                                                                          |
+| children   | string   |          |          | Use to add the content. Takes in text content only                                                                 | 
+| className  | string   |          |          | Use to declare custom class name. This will be added on the outmost div                                            | 
+| picogram   | node     |          |          | Takes in a custom pictogram node from `carbon components` that is imported to the top of the page.                 |
+| title      | string   |          |          | Prop used for a short title                                                                                        |  
+| titleType  | string   |          |          | Optional prop used for type options - default is `expressive-04`. Options are: `expressive-03` and `expressive-02` |
```
alisonjoseph, line 135: What is the default title type? Is there guidance on when to use which type? (probably a design @DianaStanciulescu question)

alisonjoseph: I also wonder if it would be better to call it title size and have a small, medium and large? 🤔 @sstrubberg thoughts here? Do we have any text size props in any other components? (I don't think so, but not sure)

packages/example/src/pages/components/ExpressiveList.mdx
```diff
@@ -0,0 +1,135 @@
...
+    background="true">
+  Lorem ipsum tempor incididunt ut labore et dolore magna aliqua. Quam lacus suspendisse faucibus interdum posuere lorem. At quis risus sed lorem ipsum vulputate.
+  </ExpressiveList>
+</div>
+```
+
+<Title>With pictograms</Title>
+
+```mdx path=components/ExpressiveList/ExpressiveList.js src=https://github.com/carbon-design-system/gatsby-theme-carbon/tree/main/packages/gatsby-theme-carbon/src/components/ExpressiveList
+<div className="expressive-content-list-group">
```
alisonjoseph, line 98: I think I would add this to the example code, otherwise I could easily see someone forget to add it and wonder why it wasn't showing up 😄 

`import { Connect } from '@carbon/pictograms-react';`

packages/gatsby-theme-carbon/src/components/ExpressiveList/ExpressiveList.module.scss
```diff
@@ -0,0 +1,47 @@
...
+.background {
+  background: $ui-01;
+}
+
+.title {
+  @include carbon--type-style('expressive-heading-04', true);
+
+  padding-right: 2rem;
+
+  @media (max-width: 847px) {
```
alisonjoseph, line 20: curious what this is for?

packages/example/src/pages/components/ExpressiveList.mdx
DianaStanciulescu: > What is the default title type? Is there guidance on when to use which type? (probably a design @DianaStanciulescu question)

The title size depends on the amount of text and also on how the rest of the page looks like and where the list is inserted in the page layout. But I do not have exact specs for this, it;s left to each designer to choose. 

If we were to have a default, I think expressive heading 03 (medium) makes the most sense since it's the middle ground between the two.  I would like @jeanservaas 's opinion here too.

sstrubberg: I like `titleType` because the first thing that pops in my head is the [type tokens](https://carbon-elements.netlify.app/type/examples/preview/).

sstrubberg:
image

Thoughts?

packages/example/src/pages/components/ExpressiveList.mdx
```diff
@@ -0,0 +1,135 @@
+---
+title: ExpressiveList
+description: Usage instructions for the FeatureCard component
+---
+
+import { Connect } from '@carbon/pictograms-react';
+
+<PageDescription>
+
+The `<ExpressiveList>` component is meant to be used with the `expressive-content-list-group` className when grouped together. This component is only intended for paragraph content and not meant to nest other components within it. `ExpressiveListContainer` coming soon. This can be used with a pictogram by importing a pictogram from our library and passing it into the `picogram` prop as a node.
```
sstrubberg, line 10:
```suggestion
The `<ExpressiveList>` component is meant to be used with the `expressive-content-list-group` className when grouped together. This component is only intended for paragraph content and not meant to nest other components within it. `ExpressiveListContainer` coming soon. This can be used with a pictogram by importing a pictogram from our library and passing it into the `pictogram` prop as a node.
```

sstrubberg (changes requested):
Really solid work here!

DianaStanciulescu:
> image
> 
> Thoughts?

Actually, it should work with imagery as well, and even cards, resource cards, etc. (these are patterns already in use, but they have been custom coded) 

Screenshot 2021-12-13 at 18 03 47

packages/gatsby-theme-carbon/src/components/ExpressiveList/ExpressiveList.module.scss
aledavila: I was trying to adjust title with the spacing at the breakpoints. At that point its wraps the title. But ill probably just do below at breakpoint lg

aledavila:
I'm just wondering if they should all be the same spacing. If not they should be separate components. Heres the instances: 

Where anything can be in the component:
Screen Shot 2021-12-13 at 11 58 32 AM

When its only content: 
Screen Shot 2021-12-13 at 11 58 25 AM

aledavila:
Seperate container component: https://github.com/carbon-design-system/gatsby-theme-carbon/pull/1198

jeanservaas (commented):
Hey Ale, 

@DianaStanciulescu will review this tomorrow but I just have a quick comment

The components are responding well... the way their displayed on the gatsby site might be a little confusing though in that you're not supposed to use list items with a smaller title in the same block as list items with a larger title...

So here it's just a little confusing because you have two list items that are the same and then one with a smaller title. I'm wondering if it would be clearer to show them as separate examples... even though it's just a style variation on the same compoent. Any thoughts on this @DianaStanciulescu or @mjabbink ?

image

DianaStanciulescu:
Hi @aledavila ! regarding this component, I see there may be some confusion. For what is now shown on the gatsby site preview under expressive list—the white background version—it is correct that " it is not meant to nest other components within it".  and they look good, and there is no other variation needed. the only thing is that the possible title sizes are exactly the same as H2, H3, H4, but if we want to call them "expressive-01", 2...in the props table, that is fine. 


But in the original spec file, there was another set of expressive list specs: without a white background, using rules, a title, and various content types on the right: paragraphs, images, gallery, cards, etc. In this case, the expressive list would allow for this mix of content types, or components; the spacing will be the normal spacing these components use, apart from the space between the rule and the first piece of content which is always 24px; after that, if you insert an image it would use 24px, if you insert a gallery, 32, a paragraph 24, a card 24, etc. I thought we were talking about this one.

xlg-1312px-12 column


@jeanservaas showing examples separately would be best, but if the page gets too long, we can show the 3 different title sizes, and just add guidance text saying that they should not be mixed.

DianaStanciulescu:
> @DianaStanciulescu @sstrubberg @alisonjoseph I'm just wondering if they should all be the same spacing. If not they should be separate components. Heres the instances:
> 
> Where anything can be in the component: Screen Shot 2021-12-13 at 11 58 32 AM
> 
> When its only content: Screen Shot 2021-12-13 at 11 58 25 AM

Thinking about it now, I think they should be separate components. They are pretty different and we would have too many variables otherwise.

sstrubberg:
Diana, Ale, Alison, and myself are going to sync on this in the new year to close it out :)

mjabbink:
@DianaStanciulescu and @jeanservaas I like the idea of showing some static alternative options this component can be used as I see in these examples https://github.com/carbon-design-system/gatsby-theme-carbon/pull/1197#issuecomment-993417839

It was not clear to me is those were to be additional components or are they variations/layouts doable with the one components.

DianaStanciulescu:
> @DianaStanciulescu and @jeanservaas I like the idea of showing some static alternative options this component can be used as I see in these examples [#1197 (comment)](https://github.com/carbon-design-system/gatsby-theme-carbon/pull/1197#issuecomment-993417839)
> 
> It was not clear to me is those were to be additional components or are they variations/layouts doable with the one components.

those should be possible with one component in which one nests other existing components. we can show them as examples in the guidance, sure. And I can add them to the gatsby kit once we decide next year how they will be coded.

DianaStanciulescu:
hi! based on our conversation:
- the background is always on
- there should be no bottom rule for the last element
- double-check the pictogram specs

aledavila:
Bump @sstrubberg @DianaStanciulescu @jeanservaas ready to be checked again :) thanks

DianaStanciulescu:
@aledavila  looks great to me!

alisonjoseph (approved):
This LGTM! Only question I had is if we want to show the various heading sizes in the demo and add any guidance on when to use them? Probably a question for @DianaStanciulescu or @jeanservaas

sstrubberg (approved):

DianaStanciulescu:
We can show heading sizes in the demo if we show it in groups - one group of rows with h2, one with h3... and maybe that takes too much space?

aledavila:
@DianaStanciulescu @alisonjoseph Perhaps just some guidance on when to use what is fine. Adding them to the demo might make it really long. Do you have any suggestions to what add.

DianaStanciulescu:
@aledavila we don't really have guidance here, it's pretty open + we have done no research to come up with strict rules. so we can leave this as it is and add guidance later if needed.

some thoughts for considering later if ever:
- make the titles smaller or equal to the section title they are nested under (if something sits under an H3, do not use H2 for titles)
- choose hs depending on character count so titles stay on 2 or 3 lines max.

mjabbink:
@aledavila and @DianaStanciulescu having a few more demos will not make the page too long. It’s better to show the few @DianaStanciulescu is mentioning visually than not show it. @DianaStanciulescu please provide a sketch for what you think is needed on the page. These could also be images rather than code (demos) just to demonstrate those items you mentioned.

aledavila:
@DianaStanciulescu I've gone ahead and added just 1 example showing each of the title types. Please let me know if this is ready Thank you

DianaStanciulescu:
@aledavila great idea! Can you please do the same for the version with pictogram? Thanks!

aledavila:
@DianaStanciulescu done

DianaStanciulescu:
@aledavila looks clear to me!

alisonjoseph:
Nice! This looks ready to merge 🥳

PR #1203, opened 2022-01-07
chore(theme): update to 4.4

sstrubberg:
updated Theme and Example to Gatsby 4.4 for the `next` branch. Once merged, I'll publish to npm under the `next` tag so we can test it against the Carbon site and IDL.

Also merged latest changes from `main`.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,1 +1,1 @@
-* @carbon-design-system/developers-auxiliary
+* @carbon-design-system/developers-carbon-brand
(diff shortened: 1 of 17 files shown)


alisonjoseph (approved):
LGTM from what I can see, lets publish and test against Carbon/IDL. Any breaking changes you can think of, I haven't looked at what's new in v4 yet.

sstrubberg:
> LGTM from what I can see, lets publish and test against Carbon/IDL. Any breaking changes you can think of, I haven't looked at what's new in v4 yet.

Vince mentioned he ran into some weird issues, but was unclear if it was cache'ing related. 🤞

PR #1212, opened 2022-01-25
fix: update the references to the new homepage

rodet:
I found out late last year that most of the URLs in the code base still referenced the old homepage [https://gatsby-theme-carbon.now.sh](https://gatsby-theme-carbon.now.sh). The current and long-term URL is now [https://gatsby.carbondesignsystem.com/](https://gatsby.carbondesignsystem.com/), so I'm updating all references to the new URL. One of the advantages of that fix is that, as the fonts are loaded over the host too, the font loading will be faster as the redirect from `gatsby-theme-carbon.now.sh` to `gatsby-theme-carbon.vercel.app` will be avoided.

#### Changelog

**Changed**

- References to the old URL.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,8 +7,8 @@ The Gatsby Carbon theme includes all the configuration you need to build a beaut
 ## Resources
 
 - [Contribution guidelines](.github/CONTRIBUTING.md)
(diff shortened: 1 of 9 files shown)


alisonjoseph (approved):
Thanks for the PR!

sstrubberg (approved):

PR #1220, opened 2022-02-09
fix: do dont component updates

### LINKED ISSUE #1217: Wrong color on DontDoExample page
## Detailed description

> Describe in detail the issue you’re having.
Linked [here](https://gatsby-theme-carbon.vercel.app/components/DoDontExample/) is the Don't Do example that is the wrong color. The example is using a darker green (#24a148) than what should be used (reference proper green [here](https://gatsby-theme-carbon.vercel.app/components/DoDontRow)). 

The expectation is a lighter green on the Do Don't example given

Chrome

## Additional information
Attached is a screenshot of the darker green in use
Screen Shot 2022-02-02 at 5 25 49 PM

alisonjoseph:
Closes #1217 

- color fix for `DoDontExample`
- adds deprecation warning for  `DoDontExample`
- ports over aspect ratio updates to the correct `DoDont` component

### Testing
- ensure color green is correct for `DoDontExample`
- ensure aspect ratios are correct for `DoDont`

### DIFF
--- a/packages/gatsby-theme-carbon/src/components/DoDontExample/do-dont-example.scss
+++ b/packages/gatsby-theme-carbon/src/components/DoDontExample/do-dont-example.scss
@@ -100,7 +100,7 @@
 }
 
 .#{$prefix}--example--correct .#{$prefix}--example-card::before {
-  background: $support-02;
+  background: #24a148;
 }
 
 .#{$prefix}--example--incorrect .#{$prefix}--example-card::before {
@@ -118,7 +118,7 @@
 }
 
 .#{$prefix}--example__icon--correct {
-  fill: $support-02;
+  fill: #24a148;
 }
 
 .#{$prefix}--example__icon--incorrect {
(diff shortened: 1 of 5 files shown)


packages/gatsby-theme-carbon/src/components/DoDontExample/do-dont-example.scss
```diff
@@ -100,7 +100,7 @@
 }
 
 .#{$prefix}--example--correct .#{$prefix}--example-card::before {
-  background: $support-02;
+  background: #24a148;
```
sstrubberg, line 103: Is there a reason we are hard-coding the colors instead of using a token?

sstrubberg (changes requested):

packages/gatsby-theme-carbon/src/components/DoDontExample/do-dont-example.scss
alisonjoseph: Not sure, this was the color requested in the issue, the token was displaying the wrong color. 
This component is also being deprecated in favor of the newer one, so I didn't spend any time trying to figure out the token issue.

alisonjoseph:
@sstrubberg I updated it to use the color token.

sstrubberg (approved):

PR #1229, opened 2022-03-07
Remove all usages of carbon components on `next`

tay1orjones:
And provide theme resources globally


tay1orjones:
I can cut another prerelease once this is in.

packages/gatsby-theme-carbon/src/styles/internal/resources.scss
```diff
@@ -13,5 +13,8 @@
 @use '@carbon/react/scss/type' as *;
 // $standard-easing $transition-x
 @use '@carbon/react/scss/motion' as *;
+// @include visually-hidden;
+@use '@carbon/react/scss/utilities/visually-hidden' as *;
```
sstrubberg, line 17: Should it be importing from `/styles` instead of `/react`?

sstrubberg (changes requested):
Should we also take this opportunity to bring Gatsby up from `4.4` to `4.9.2`?

tay1orjones:
@sstrubberg I'll try it! I think v4.7 has something like a 30% page creation/build time improvement, so def worth it if it isn't a huge headache.

packages/gatsby-theme-carbon/src/styles/internal/resources.scss
tay1orjones: @sstrubberg I stand corrected. In this case the visually-hidden helper isn't provided through `@carbon/react` yet. I'll fix this to use the hidden styling directly instead of having to rely on `@carbon/styles`.

sstrubberg (approved):

PR #1234, opened 2022-03-23
chore(project): update to carbon 10.56.0, correct style import order

tay1orjones:
This updates the carbon dependencies, but also corrects the import order so modules can be properly configured

~Going to leave in draft until we can update this with the v11 RC~


tay1orjones:
cc @aledavila FYI

tay1orjones:
Updated to 10.56.0

PR #1242, opened 2022-05-18
fix(grid): add _grid.scss and move over flex grid for precedence

dakahn:
ref https://github.com/carbon-design-system/carbon-website/issues/2884

Grid styles were being overwritten which was resulting in a busted StatusIndicator table
image

This PR moves grid styles into their own file and requires that file above the other mixins so those styles take precedence and get brought in properly 🤠 

## Testing

Make sure it builds and compiles as expected 👍🏾

### DIFF
--- a/packages/gatsby-theme-carbon/src/styles/internal/_global.scss
+++ b/packages/gatsby-theme-carbon/src/styles/internal/_global.scss
@@ -1,4 +1,4 @@
-@use '@carbon/grid/scss/flex-grid';
+@use 'grid';
 @use '@carbon/react/scss/type';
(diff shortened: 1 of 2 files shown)


tay1orjones (approved):

tay1orjones:
@dakahn let me know if you want to walk through releasing this together. It'll need to be published as a prerelease to the `next` tag once merged in.

joshblack (approved):

tay1orjones:
@dakahn I just published this in https://github.com/carbon-design-system/gatsby-theme-carbon/releases/tag/v3.0.0-next.11

PR #1246, opened 2022-06-16
fix(markdown): ensure links within lists do not have padding

tay1orjones:
This fixes the bug blocking https://github.com/carbon-design-system/carbon-website/pull/2975

#### Changelog

**Changed**

- Ensures that links placed as the second/third/fourth/etc sibling within a list item don't have unnecessary padding.
- Update sass version
- Formatting within the demo page

**Removed**

Demo page list section should render the new demo text with no extra vertical padding around links


tay1orjones:
👍 

image

PR #1376, opened 2023-11-29
feat: add option for 3 across ResourceCard group

### LINKED ISSUE #1375: Resource Card Group enhancement - add 3 across option
From @mjabbink 

> It has always been the intent to get resource cards to be go across the page in 2 or 3. However it was developed to be only 2. 
> 
> The Carbon example shows how 3 would be a more efficient solution as well as look better. 
> https://carbondesignsystem.com/#other-resources
> 
> In the Carbon for ibm.com since there’s only three it wold also benefit from being 3 across (simplified)
> https://ibmdotcom-library.s3-web.us-south.cloud-object-storage.appdomain.cloud/deploy-previews/1688/

Current design - only option is 2 across
Screenshot 2023-11-29 at 8 55 24 AM

Option to be added (an example of 3 across)

Screenshot 2023-11-29 at 8 59 00 AM

Current code that creates correct borders and span only two across
```
<Row className="resource-card-group">
  <Column colMd={4} colLg={4} noGutterSm>
    <ResourceCard...... 
```

One suggestion is adding a new component, `CardGroup`, similar to what we did in Platform, and then add a prop to choose how many show across, with the default being 2. 

Possibly `cardSpan`? or `cardsPerRow`?
```
<CardGroup cardSpan={3}>
 <Column colMd={4} colLg={4} noGutterSm>
    <ResourceCard...... />
```

OR a simple option is to just add an additional class-name to add correct styling
```
<Row className="resource-card-group-3-across">
  <Column colMd={4} colLg={4} noGutterSm>
    <ResourceCard...... 
```

alisonjoseph:
Closes #1375

#### Changelog

**New**

- Add new classname for 3 across card group.

### DIFF
--- a/packages/example/src/pages/components/ResourceCard.mdx
+++ b/packages/example/src/pages/components/ResourceCard.mdx
@@ -6,12 +6,15 @@ description: Usage instructions for the ResourceCard component
 <PageDescription>
 
 The `<ResourceCard>` component should be wrapped with a `<Column>` inside of
-`<Row className="resource-card-group">`. This allows the correct border
-placement between a group of cards.
+`<Row className="resource-card-group">` or
+`<Row className="resource-card-group-3-across">`. This allows the correct border
+placement between a group of cards spanning 2 or 3 across.
 
 </PageDescription>
 
 ## Example
+
+### Group two across
 
 <Row className="resource-card-group">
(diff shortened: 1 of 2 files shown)


andreancardona (approved):

thyhmdo (approved):
Looks great!

Issue #1391, opened 2024-01-08
Update Plex to latest


alisonjoseph:
Need to update to latest Plex to fix alignment issues that were fixed in v6.3 https://github.com/IBM/plex/issues/377

Currently Plex is being self hosted
https://github.com/carbon-design-system/gatsby-theme-carbon/tree/main/fonts
https://github.com/carbon-design-system/gatsby-theme-carbon/blob/main/packages/gatsby-theme-carbon/src/styles/internal/_plex.scss

ideally we could pull from the cdn, however it doesn't exist there yet. https://github.ibm.com/webstandards/plex-cdn

PR #1392, opened 2024-01-08
feat: add font files for plex 6.3 release

alisonjoseph:
Ref #1391

this just adds the new font files. I wasn't able to simply replace them as they seem to be named differently. 

there will be separate PR to update
https://github.com/carbon-design-system/gatsby-theme-carbon/blob/main/packages/gatsby-theme-carbon/src/styles/internal/_plex.scss

and 

https://github.com/carbon-design-system/gatsby-theme-carbon/blob/main/packages/gatsby-theme-carbon/gatsby-ssr.js#L35

once fonts files are live

*note: variable fonts aren't served via cdn yet


tay1orjones (approved):
It's so bizarre how we serve these. IIUC the theme demo/docs site deployed to vercel is essentially a CDN for plex on all sites using gatsby-theme-carbon. 😵‍💫

PR #1410, opened 2024-02-03
fix(scss): import + other minor fixes

muditjuneja:
This PR aims at fixing a few things that are causing build failures which include -

- Mdx updated format
- Fix a few imports as per the upgrades

This PR should be merged to the next branch.


muditjuneja:
@eng618 Can you review this PR?

packages/gatsby-theme-carbon/src/components/Code/Code.js
```diff
@@ -50,7 +52,7 @@ const Code = ({ children, className: classNameProp = '', path, src }) => {
         {children}
       </PathRow>
       <Highlight
-        {...defaultProps}
+        // {...defaultProps}
```
eng618, line 55: Why did you comment this out?

packages/gatsby-theme-carbon/src/components/LeftNav/LeftNavItem.js
```diff
@@ -2,7 +2,7 @@ import React, { useContext } from 'react';
 import { Link } from 'gatsby';
 import { Location } from '@reach/router';
 import cx from 'classnames';
-import useNetwork from 'react-use/lib/useNetwork';
+import {useNetworkState} from 'react-use';
```
eng618, line 5: A bit nit picky here, but make sure to apply formatting (i.e. spaces round this)

packages/gatsby-theme-carbon/src/components/Code/Code.js
```diff
@@ -1,5 +1,5 @@
 import React, { useState } from 'react';
-import Highlight, { defaultProps } from 'prism-react-renderer';
+import { Highlight, themes } from "prism-react-renderer"
```
eng618, line 2: Single quotes here.

eng618: Looks like you are importing `themes` here but not using it anywhere?

eng618 (changes requested):
Looks mostly good, and while it compiles locally, we are picking up the wrong theme in the pages. I feel like this is because the tokens are not being picked up or imported correctly. 

image

vs what it should be: 

image

packages/gatsby-theme-carbon/src/components/Code/Code.js
muditjuneja: Brought back the defaultProps in this component.

packages/gatsby-theme-carbon/src/components/LeftNav/LeftNavItem.js
muditjuneja: Applied lint fix.

muditjuneja:
@eng618 `useMetadata` was not being called in DefaultLayout and it was not able to supply the theme variables. Fixed it by calling the method.
Another thing that I am noticing is that the Code component is not working as expected and we would need to look more into that. Other than this, build looks good.

packages/gatsby-theme-carbon/src/components/Code/Code.js
```diff
@@ -23,25 +23,27 @@ const Code = ({ children, className: classNameProp = '', path, src }) => {
   const language = classNameProp.replace(/language-/, '').replace('mdx', 'jsx');
 
   const removeTrailingEmptyLine = (lines) => {
-    const [lastLine] = lines.splice(-1);
-    if (lastLine[0].empty) {
-      return lines;
+    if (lines && lines.length) {
```
muditjuneja, line 26: These checks are almost random that I applied to make the build work as this component could not find lines in a few pages. It would make sense to look into this as a separate PR.

muditjuneja:
Also, 
Text defined like below - 

The `<ResourceCard>` component should be wrapped with a `<Column>` inside of
`<Row className="resource-card-group">` or
`<Row className="resource-card-group-3-across">`. This allows the correct border
placement between a group of cards spanning 2 or 3 across.

is being rendered differently now. See the diff below 

Screenshot 2024-02-04 at 1 47 12 AM
Screenshot 2024-02-04 at 1 47 25 AM

packages/gatsby-theme-carbon/src/components/Code/Code.js
```diff
@@ -17,33 +17,46 @@ import useMetadata from '../../util/hooks/useMetadata';
...
 
     if (shouldShowMore) {
       return withoutTrailingEmpty;
     }
 
-    return withoutTrailingEmpty.slice(0, 9);
+    return withoutTrailingEmpty ? withoutTrailingEmpty.slice(0, 9) : [];
   };
 
+  if (isInlineCode) {
```
muditjuneja, line 57: With this version, inline code are being rendered as pre only and not as <code></code>. This has fixed that. (Maybe temporarily)

packages/example/src/pages/components/markdown.mdx
```diff
@@ -145,7 +145,7 @@ recommend using `h2` tags for section headings within your content.
 [I’m a markdown link with title](https://www.carbondesignsystem.com 'Carbon’s Homepage')
 
 URLs and URLs in angle brackets will automatically get turned into links.
-http://www.example.com or <http://www.example.com> and sometimes example.com
+http://www.example.com or \<http://www.example.com> and sometimes example.com
```
eng618, line 148:
```suggestion
`http://www.example.com` or `<http://www.example.com>` and sometimes example.com
```

eng618: Lets just add these as code blocks so they are not interpreted, as this is technically what it should look in markdown code anyways.

packages/gatsby-theme-carbon/src/components/Code/Code.js
eng618: Add a `// TODO: message...` for this as well, so we can remember to look into and verify this file all together.

eng618 (changes requested):
Just a couple minor things, then we can get this merged in.

PR #1412, opened 2024-02-06
Revert "fix(scss): import + other minor fixes"

alisonjoseph:
Reverts carbon-design-system/gatsby-theme-carbon#1410

Hey @eng618 there are some broken styles as a result of this PR, I'm reverting. Resource card group styling for one, but haven't had time to look into if there is anything else or what the cause is.


eng618:
@alisonjoseph how did you come to the conclusion it was this PR? As this should only affect the next branch?

alisonjoseph:
@eng618 oh, I missed that. then this is the wrong PR

PR #1417, opened 2024-02-09
fix: mdx tables + port gatsby core to ESM

muditjuneja:
This PR has - 
1. Mdx tables fix - With migrating to v3 of mdx, the gfm capabilities like tables, etc have been deprecated and the suggestion is to use another module named `remark-gfm`
2. Porting core gatsby files to ES modules - gatsby-config.js + gatsby-node.js


packages/gatsby-theme-carbon/gatsby-node.mjs
```diff
@@ -0,0 +1,120 @@
+import path from 'path';
```
muditjuneja, line 1: I am not sure why git renamed gatsby-config.js to gatsby-config.mjs and is adding this as a new file after removing the older one.

muditjuneja:
@eng618 Can you review this PR?

packages/gatsby-theme-carbon/gatsby-node.mjs
eng618: You probably need to set `"type:" "module"` in the **package.json** so that the entire project defaults to modules instead of CommonJS. Give that a try and see if you can rename the files back to just .js

muditjuneja: To be honest I was just trying to do that only. But I am getting this - 
`
 ERROR #10123  API.CONFIG.LOADING

We encountered an error while trying to load your site's gatsby-config. Please fix the error and try again.



  ReferenceError: require is not defined in ES module scope, you can use import instead
  
  - gatsby-config.mjs:1 
    file:///Users/mjuneja/Work/mine/OS/gatsby-theme-carbon/packages/gatsby-theme-carbon/gatsby-config.mjs:1:14
  
  - module_job:194 ModuleJob.run
    node:internal/modules/esm/module_job:194:25
  
`

packages/gatsby-theme-carbon/gatsby-config.js
```diff
@@ -111,7 +117,7 @@ module.exports = (themeOptions) => {
             ...gatsbyRemarkPlugins,
           ],
           mdxOptions: {
-            remarkPlugins,
+            remarkPlugins: [...remarkPlugins, remarkGfm],
```
eng618, line 120:
```suggestion
            remarkPlugins: [remarkGfm, ...remarkPlugins],
```

I would flip these in case the user of the theme want to overwrite some settings or pass in their own version of the plugin they can do so.

eng618 (changes requested):
One minor adjustment, other than that this looks great!

eng618 (approved):

muditjuneja:
Closing this as another PR will be merged for this - https://github.com/carbon-design-system/gatsby-theme-carbon/pull/1418#issuecomment-1939135900

Issue #868, opened 2020-06-03
MDX code block imporvements


ENG618:
### Summary

Code blocks should have the option to not shrink (i.e. remove the `Show More/Less` and leave the full text),

Just development

### Justification

Makes our tutorials harder to digest

### Desired UX and success metrics

The user could see the entire code block presented to them, so there is not extra steps to click `Show More/Less`.   

Also in some cases when the code block is in a nested list and the code box is to small (because #832) the `Show More/Less` button actually covers some of the text needing to be rendered. (so possible a bug 🐛 as well, maybe some more padding for the button 🤷‍♂️)

### Must-have functionality

prop to disable virticle overflow (`Show More/Less)`

### Specific timeline issues / requests

ASAP, we have already ran into issues with multiple people hitting the same issue when going throug this tutorial.

### Available extra resources

I could look into it with a bit of guidance.

aharbis:
In addition to this, I think it would be beneficial to offer a word-wrap toggle, so that horizontal scrolling could be disabled. For example, it's not a great user experience to represent a very long log message in a block this:

```
Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat.
```

Nor does putting in single back-tick (in-line) code look good:

`Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat.`

If the code block could support optional word-wrap or horizontal scroll, that would allow us to better represent wide data (such as log messages) without the user needing to scroll to see the whole message, and never being able to see more than 20% of the message.

PR #1422, opened 2024-02-20
Fix: code blocks in mdx + issues/868

muditjuneja:
This PR fixes the following - 

- Show more button in Code blocks
- Added a prop to enable/disable Show more button in Code blocls
- Added text wrap to avoid horizontal scroll
- Added a new package that is traversing hast and adding a new property named metaData which is being passed as a prop

Should close - https://github.com/carbon-design-system/gatsby-theme-carbon/issues/868

### DIFF
--- a/packages/gatsby-theme-carbon/gatsby-config.mjs
+++ b/packages/gatsby-theme-carbon/gatsby-config.mjs
@@ -9,6 +9,16 @@ import remarkGfm from 'remark-gfm';
 import { fileURLToPath } from 'url';
 import defaultLunrOptions from './config/lunr-options.mjs';
 
+/*
+  This is a rehype plugin that adds support for metadata to the fenced code block
+  For eg: 
+  ```jsx path=/directory/file.mdx src=https://gatsby.carbondesignsystem.com
+  const a = 10;
+  ```
+  A metaData prop of format path=/directory/file.mdx src=https://gatsby.carbondesignsystem.com is added to the code block
+*/
+import rehypeAddCodeMetaData from 'rehype-mdx-fenced-code-meta-support';
+
 const __dirname = dirname(fileURLToPath(import.meta.url));
 const carbonThemes = {
   white: './src/styles/internal/white.scss',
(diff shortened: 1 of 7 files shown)


packages/gatsby-theme-carbon/src/components/Code/Code.js
```diff
@@ -14,7 +14,13 @@ import Sidebar from './Sidebar';
 
 import useMetadata from '../../util/hooks/useMetadata';
 
-const Code = ({ children, className: classNameProp = '', path, src }) => {
+const Code = ({
+  children,
+  className: classNameProp = '',
+  path,
+  src,
+  hideCode = false,
```
muditjuneja, line 22: With this, by default, we are not doing the "Show more" in code blocks.

eng618 (changes requested):
The changes look really good, the only thing I would like to see is the examples updated to reflect this change. Especially adding an example for the **Show More** prop. This would be in the **packages/example/src/pages/components/code-blocks.mdx** file.

packages/gatsby-theme-carbon/gatsby-config.mjs
```diff
@@ -9,6 +9,16 @@ import remarkGfm from 'remark-gfm';
...
 
+/*
+  This is a rehype plugin that adds support for metadata to the fenced code block
+  For eg: 
+  ```jsx path=/directory/file.mdx src=https://gatsby.carbondesignsystem.com
+  const a = 10;
+  ```
+  A metaData prop of format path=/directory/file.mdx src=https://gatsby.carbondesignsystem.com is added to the code block
+*/
+import rehypeAddCodeMetaData from 'rehype-mdx-fenced-code-meta-support';
```
muditjuneja, line 20: Created a rehype package which is adding a new prop of named metaData which is being passed in Code.js.

Here is the dirty package - https://www.npmjs.com/package/rehype-mdx-fenced-code-meta-support

eng618:
Lets add some padding to the show more button so it doesn’t overlap the content.

image

eng618 (approved):
Approving now, but would be good to add the padding to the show more button before we merge this in.

muditjuneja:
> Lets add some padding to the show more button so it doesn’t overlap the content.
> 
> image

I will do that in another follow up PR. Merging this one now.

muditjuneja:
> > Lets add some padding to the show more button so it doesn’t overlap the content.
> > image
> 
> I will do that in another follow up PR. Merging this one now.

I guess I can't merge. I will make that small change.

Issue #1425, opened 2024-02-28
Gatsby v5: Review `next` branch and make issues


alisonjoseph:

alisonjoseph:
Some visual discrepencies
`next` on the left, `main` on the right

- [ ] Nested list styling is off

<img width="1012" alt="Screenshot 2024-02-29 at 12 35 00 PM" src="https://github.com/carbon-design-system/gatsby-theme-carbon/assets/2753488/a2705a60-cd18-400a-9a20-423bf96510a8">

- [ ] Spacing above images is missing (looks to be wrapped in an additional span tag)

<img width="912" alt="Screenshot 2024-02-29 at 12 36 31 PM" src="https://github.com/carbon-design-system/gatsby-theme-carbon/assets/2753488/0de7c4c2-1317-4c32-96e8-863efe2877af">

alisonjoseph:
Remaining issues are tracked in 
https://github.com/carbon-design-system/gatsby-theme-carbon/issues/1439
https://github.com/carbon-design-system/gatsby-theme-carbon/issues/1440

PR #1438, opened 2024-03-08
fix: nested list styling for v5

alisonjoseph:
Ref #1425

Updates how nested lists are rendered 
```
'li.ul': (props) => <Ul nested {...props} />,
'li.ol': (props) => <Ol nested {...props} />,
```
  
no longer works with latest versions, so used the solution we have over in Platform to render the correct classnames for nested lists. This also allows us to remove a lot of the custom styling we had and just use the base Carbon list styles. 

## Testing

Ensure spacing looks the same as live site for markdown lists.


eng618 (approved):
Looks good

PR #1443, opened 2024-03-13
docs: code blocks docs page

alisonjoseph:

PR #1444, opened 2024-03-13
Fix(hideCode) - Checking for stringified boolean

muditjuneja:
This PR fixes how hideCode prop is handled as a string value.


alisonjoseph:
@muditjuneja I pulled your change into https://github.com/carbon-design-system/gatsby-theme-carbon/pull/1443 FYI CI will fail if you don't use the correct [commit conventions](https://github.com/carbon-design-system/carbon/blob/main/docs/developer-handbook.md#type).

muditjuneja:
My bad, Sometimes I forget that commit lint is present here.

PR #1445, opened 2024-03-16
refactor: change hideCode prop to showAll prop

muditjuneja:
This PR aims at changing the hideCode prop to showAll prop as the naming makes more sense with showAll.


packages/gatsby-theme-carbon/src/components/Code/Code.js
```diff
@@ -33,7 +33,16 @@ const Code = ({ children, className: classNameProp = '', metaData }) => {
     if (metaData) {
       const metaDataObject = metaData.split(' ').reduce((obj, item) => {
         const [key, value] = item.split('=');
-        obj[key] = value;
+
+        // Checking for boolean values coming as string or not coming like in showAll prop
```
muditjuneja, line 37: I believe this entire logic should be offloaded to the package. I will see if I can get that working but not in this PR.

eng618 (approved):

PR #1450, opened 2024-03-20
fix: warnings

alisonjoseph:
Closes #

{{short description}}

#### Changelog

**New**

- {{new thing}}

**Changed**

- {{change thing}}

**Removed**

- {{removed thing}}

### DIFF
--- a/packages/gatsby-theme-carbon/gatsby-node.js
+++ b/packages/gatsby-theme-carbon/gatsby-node.js
@@ -1,9 +1,21 @@
-import path from 'path';
-import fs from 'fs';
-import * as mkdirp from 'mkdirp';
-import startCase from 'lodash.startcase';
+const path = require('path');
+const fs = require('fs');
+const mkdirp = require('mkdirp');
+const startCase = require('lodash.startcase');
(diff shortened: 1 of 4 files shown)


alisonjoseph:
@eng618 thoughts on this to silence all those warnings, or worth digging in to figure out if they are actually an issue. I think it we may be able to safely ignore? https://robertmarshall.dev/blog/fix-warn-chunk-commons-mini-css-extract-plugin-error-in-gatsby-js/

eng618:
> @eng618 thoughts on this to silence all those warnings, or worth digging in to figure out if they are actually an issue. I think it we may be able to safely ignore? https://robertmarshall.dev/blog/fix-warn-chunk-commons-mini-css-extract-plugin-error-in-gatsby-js/
> 
> I did notice there is a difference with this version on the precedence of site styles vs. Carbon styles, but couldn't figure out exactly why.

@alisonjoseph Yeah, I think it's safe to filter those out also 🙌🏼 

The only question I have is there a reason we need to move away from js modules (**.mjs**) file back to common js (**.js**) file?

alisonjoseph:
@eng618 oops, I was just testing things, swapped it back.

eng618 (approved):

PR #1502, opened 2024-08-08
[Draft]: Left nav tree

### LINKED ISSUE #1363: Add side nav tree
### Summary

This is an extension of https://github.com/carbon-design-system/gatsby-theme-carbon/issues/1362 that should be completed together.

The current side nav is limited by one level of nesting. This has resulted in a cluttered side nav, and an overly aggressive use of page tabs.

The Carbon core team built an entirely new way to do this in Carbon Platform by taking the (now stable) `TreeView` component (part of Carbon v11):

https://react.carbondesignsystem.com/?path=/docs/components-treeview--overview

And then theming and customizing it to work as a side nav with multiple levels of nesting. The Carbon Platform doesn't have an example of level two nesting, but here's an example from the design spec:

image

### Justification

This will allow much needed UX and IA improvements for the various Carbon local system (PAL) websites that are build with the Carbon Gatsby Theme.

Match the coded `TreeNav` component used in Carbon Platform: https://next.carbondesignsystem.com

Here's the design spec used to build Carbon Platform, in case a bug is identified in Carbon Platform: https://www.figma.com/file/UkKF0SLT3qUmGoXQFVKtXT/RL_Library_Navigation?type=design&node-id=0-1&mode=design&t=unaj7OSZxa6uc7Zp-0

### Must-have functionality

This is the current nav data model:

```yaml
- title: Menu
  pages:
    - title: Page 1
      path: /menu/Page-1
    - title: Page 2
      path: /menu/Page-2
- title: Single Page
  pages:
    - path: /single-page
```

I'm thinking we'd want to support a second level of links with a YAML structure like:

```yaml
- title: Menu
  pages:
    - title: Page 1
      path: /menu/Page-1
    - title: Page 2
      pages:
        - title: Level two page
          path: /menu/page-2/level-two-page
          - title: Another level two page
          path: /menu/page-2/another-level-two-page
- title: Single Page
  pages:
    - path: /single-page
```

Like the side nav heading issue, any changes to the nav data should included updated documentation.

Both of these issues would needed to be completed without breaking changes. Because these enhancements only add to the nav data model, that should be possible.

### Development considerations

Here is the Carbon Platform source for `NavTree` that could be lifted and shifted. Because the nav data model is different, this requires more than a copy paste job. The styling overrides could be used directly, but the logic to parse the nav data will have to be different.

https://github.com/carbon-design-system/carbon-platform/tree/main/services/web-app/components/nav-tree

Because the latest version of the Gatsby theme uses Carbon v11, this should be possible without having to update any Carbon versions.

### Specific timeline issues / requests

N/A

### Available extra resources

@glapadre might consider contributing this enhancement.

szinta:
Closes #1363 

Add support for level two nesting in left side navigation menu using carbon's TreeView component.

Screenshot of side nav with 2 level nested side nav
image

#### Changelog

**New**

- packages/gatsby-theme-carbon/src/components/LeftNav/LeftNavTree.js
- packages/gatsby-theme-carbon/src/components/LeftNav/LeftNavTree.module.scss
- packages/gatsby-theme-carbon/src/util/NavTree.js
- packages/example/src/pages/level-two -  dummy pages for the demo of 2 level nesting

**Changed**

- packages/example/src/data/nav-items.yaml
- packages/gatsby-theme-carbon/gatsby-node.mjs
- packages/gatsby-theme-carbon/src/util/NavItems.js
- packages/gatsby-theme-carbon/src/components/LeftNav/LeftNav.js
- packages/gatsby-theme-carbon/src/components/NextPrevious/NextPrevious.js

**Removed**

- none

The below listed are the pending tasks/known issues which needs to be fixed
- [ ] Next & Previous links not working properly for level 2 pages
- [ ] Style customisation as per [Figma](https://www.figma.com/design/gmlnTUu8PQtgJoiHnkgzXd/IBM-Product-PAL-IA?node-id=471-2961&t=X78Zjjpl737r0OvP-0)
- [ ] if Level 2 contains only one item render it as level 1 leaf node
- [ ] Resource links are not working with TreeNav as it is created using SideNavLink component

### DIFF
--- a/packages/example/src/data/nav-items.yaml
+++ b/packages/example/src/data/nav-items.yaml
@@ -4,6 +4,18 @@
 - title: Demo
   pages:
     - path: /demo
+- title: Level two
+  pages:
+    - title: Page one
+      path: /level-two/page-one
+    - title: Page two
+      path: /level-two/page-two
+    - title: Level three
+      pages:
+        - title: Page three
+          path: /level-two/page-three
+        - title: Page four
+          path: /level-two/page-four
 - title: Gallery
   pages:
     - path: /gallery
--- a/packages/example/src/pages/level-two/page-four.mdx
+++ b/packages/example/src/pages/level-two/page-four.mdx
@@ -1,0 +1,95 @@
+---
+title: Level two page four
+description: Quick start guide for getting acclimated with gatsby-theme-carbon
+---
+
+<PageDescription>
+
+This project is a
+[Gatsby theme](https://www.gatsbyjs.org/docs/themes/what-are-gatsby-themes/)
+that empowers developers, writers, and designers to create fast, accessible
+sites that look and feel like Carbon. Follow along to start creating your own
+site.
+
+</PageDescription>
+
+## First steps
+
+1. **Create your site** – use the gatsby CLI to bootstrap your site with the
+   starter
+
+   ```sh
+   npx gatsby new my-carbon-site https://github.com/carbon-design-system/gatsby-starter-carbon-theme
+   ```
+
+2. **Start developing** – navigate into your directory with `cd my-carbon-site`.
+
+   Start it up using one of the following snippets. You can tell which command
+   to use based on the lock file at the root of your project (`yarn.lock` for
+   yarn and `package-lock.json` for npm). For yarn, type `yarn dev` for npm,
+   you’ll use `npm run dev`.
+
+3. **Make some changes!** – open [localhost:8000](//localhost:8000) in your
+   browser to see your site running.
+
+   Each of the Items in your side bar correlates to a MDX file in your
+   `src/pages/` directory. Navigate to a site and try editing the corresponding
+   markdown file. You’ll be able to see it update live!
+
+## 👀 Watch a newbie try it out
+
+This is a [demo of a new Carbon team member spinning up a Gatsby Theme Carbon
+site](https://player.vimeo.com/video/437931932 while following the steps listed above.
+To forward through the introduction, play from min `3:50`.
+
+## 🔍 What’s in here?
+
+Lets check out the structure of our project
+
+```shell
+.
+├── LICENSE
+├── README.md
+├── gatsby-config.js
+├── node_modules
+├── package.json
+├── public
+├── src
+│   ├── data
+│   ├── gatsby-theme-carbon
+│   ├── images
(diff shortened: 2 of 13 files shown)


alisonjoseph (changes requested):
This is working great! 🥳 

Just a few questions mostly around documentation. 


If we aren’t going to add dark theme support for this we need to document that in the guide page, I’d really like to see it added though just to complete this.

If we aren’t adding dark mode support I think we should remove the extra nav pages from the example before publishing. 

Need to update this docs page explaining how it renders differently depending on levels of nesting, and how nesting is supported etc. /guides/navigation/sidebar/

If we keep the example pages I would suggest renaming/moving the Level Two/three etc pages into the demo page to keep it cleaner.
Maybe something like

Demo 
- Demo level two
- Demo
- Demo

And just leave those pages mostly blank, maybe have a paragraph stating this is a demo of a nested page and then link to the nav docs guide page.

oliviaflory:
@szinta this looks really great! 

I agree with @alisonjoseph that we could label the tree nav demo a bit better

What do you two think of either of these two options?
1. Demo level X
2. Demo tree nav level X
Screenshot 2024-08-14 at 12 10 53 PM

And a small change after we decide what the nav items should be called: the navigation label and the lead space title should always be the same text ie "Page 3" and "Page 3" or "Demo level 1" and "Demo level 1".
Screenshot 2024-08-14 at 12 13 21 PM

oliviaflory:
Thanks @alisonjoseph! @sintaibm here's a mock up of the tree demo exposed and the lead space title example:

Screenshot 2024-08-14 at 4 43 35 PM

szinta:
@oliviaflory Using identical titles for multiple leaf nodes at the same level (for instance, having all the leaf nodes under **Demo level 1** titled **Demo level 2**) is not feasible. So, I’ve adjusted the titles for those leaf nodes. Please review the updated titles and provide feedback or suggestions for improving the navigation menu titles and page description.

image

oliviaflory (approved):
@sintaibm ah yes, that makes sense why it's not feasible. I think this is probably fine, @alisonjoseph what do you think?

alisonjoseph:
Yes, that works!

alisonjoseph:
Small issue remaining with icon color in dark mode.
image

PR #1508, opened 2024-09-10
fix: resolves issue with treenav in production

davidmenendez:
fixes issue where treenav isn't responding correctly to url path in production

### DIFF
--- a/packages/gatsby-theme-carbon/src/components/LeftNav/LeftNavTree.js
+++ b/packages/gatsby-theme-carbon/src/components/LeftNav/LeftNavTree.js
@@ -67,7 +67,7 @@ const LeftNavTree = ({ items, theme }) => {
     const stripTrailingSlash = (str) =>
       str.endsWith('/') ? str.slice(0, -1) : str;
     const base = process.env.PATH_PREFIX
-      ? location.pathname.split(`${process.env.PATH_PREFIX}`)[1]
+      ? location.pathname.split(process.env.PATH_PREFIX)[1]
       : location.pathname;
     setActivePath(stripTrailingSlash(base));
   }, [location.pathname]);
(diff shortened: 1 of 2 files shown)


packages/gatsby-theme-carbon/src/components/LeftNav/LeftNavTree.js
```diff
@@ -66,7 +66,10 @@ const LeftNavTree = ({ items, theme }) => {
   useEffect(() => {
     const stripTrailingSlash = (str) =>
       str.endsWith('/') ? str.slice(0, -1) : str;
-    setActivePath(stripTrailingSlash(location.pathname));
+    const base = process.env.PATH_PREFIX
+      ? location.pathname.split(`${process.env.PATH_PREFIX}`)[1]
```
matthewgallo, line 70:
```suggestion
      ? location.pathname.split(process.env.PATH_PREFIX)[1]
```

davidmenendez: ooops duh 😅

matthewgallo (approved):
🎉

szinta (approved):

szinta:
@davidmenendez  I've identified an issue with this PR, the environment variable `process.env.PATH_PREFIX` is specific to [carbon-for-products-website](https://github.com/carbon-design-system/carbon-for-products-website/blob/63cb60ef6830de98f7402b1948ab25760a92d4b8/gatsby-config.js#L7). This means that the fix will only apply to that website or other websites with a similar configuration for setting the `pathPrefix`.

I’ll create another PR with a more generic solution. The current fix will work for carbon-for-products-website, but instead of retrieving the pathPrefix from an environment variable, we can use a GraphQL query to fetch it. This will ensure that the fix works for all Gatsby apps, regardless of the environment variable they use to set pathPrefix.

davidmenendez:
closing in favor of the solution submitted here https://github.com/carbon-design-system/gatsby-theme-carbon/pull/1509

PR #1519, opened 2024-09-18
fix: page description with tag spacing

### LINKED ISSUE #1517: [page descirption tag] need to fix a small spacing issue
After approving this recent related Gatbsy PR ([#1513](https://github.com/carbon-design-system/gatsby-theme-carbon/pull/1513)), I realized that the spacing for this is not entirely correct yet once I saw it actually live in my own PR.

I noticed this because when switching between the Usage/Style/Code/A11y tabs, the height of the description jumps down the page because of the addition of the tag. The page description needs to stay in the same place across tabs even when the tag is present and its not accounting for the space below the tag as part of the full 80px.

**Example gif of whats happening in my PR**

Sep-13-2024 11-36-43

## Spec

What I believe it is not accounting for is including the extra 16px padding below the tag as part of the full 80px.

Screenshot 2024-09-13 at 11 31 42 AM

alisonjoseph:
Closes #1517

image

#### Changelog

**Changed**

- fix spacing for PageDescription with tag

https://gatsby-theme-carbon-git-fork-alison-dacea3-carbon-design-system.vercel.app/components/PageDescription/

### DIFF
--- a/packages/gatsby-theme-carbon/src/components/PageDescription/PageDescription.module.scss
+++ b/packages/gatsby-theme-carbon/src/components/PageDescription/PageDescription.module.scss
@@ -27,4 +27,5 @@ body .page-description:has(:global(.cds--tag)) {
   top: -$spacing-08;
   margin-top: 0;
   margin-bottom: 0;
+  margin-left: 0;
 }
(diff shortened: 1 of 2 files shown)


laurenmrice (approved):

Kritvi-bhatia17 (changes requested):
Hi Alison,
The spacing from the top looks perfect now! If I’m looking at the right page, I noticed that the tag seems to have about 4px of padding on the left and right, which is why it’s not aligning with the bottom text on the left.

image

alisonjoseph:
@Kritvi-bhatia17 just pushed an update to remove the left margin from the tag

PR #1520, opened 2024-09-24
fix(leftnavtree): style updates

### LINKED ISSUE #1507: Nav divider is wrong color in dark theme
Screenshot 2024-09-05 at 7 36 53 AM

Should be gray 20 I believe (you can see [Carbon v10 website](https://v10.carbondesignsystem.com/) for correct color)

Also in light theme, we should use white background for the left nav for contrast with the gray-10 content background.

```[tasklist]
### Tasks
- [ ] Update nav divider color in dark theme
- [ ] Re-add missing nav border
- [ ] Adjust margin (saw some extra margins — probably to accommodate the banner on the demo)
- [ ] Fix theme light mode nav theme colors
```

szinta:
Closes #1507 

Fixes style issues mentioned in #1507 

Now following the below Color codes for dark and light themes

Dark theme:
```
Nav background:  #161616
Nav right Border: #393939;
Divider color: #393939
```

Light theme:
```
Nav background:  #ffffff
Nav right Border: #e0e0e0;
Divider color: #c6c6c6
```


packages/gatsby-theme-carbon/src/components/LeftNav/LeftNavTree.module.scss
```diff
@@ -7,17 +7,19 @@
 
 .container {
   height: 100%;
+  width: 100%;
+  position: absolute;
   margin: spacing.$spacing-10 0;
-  background-color: theme.$background;
+  background-color: theme.$text-inverse;
```
elycheea, line 13:
```suggestion
  background-color: theme.$layer-01;
```

Should probably be using a layer token. I think this is it?

packages/gatsby-theme-carbon/src/components/LeftNav/LeftNavTree.module.scss
elycheea, line 1: I think we’re still missing the `border-right` here. (Does gatsby-theme-carbon use logical properties? Should it be `border-inline-end`?)

https://github.com/carbon-design-system/carbon-for-products-website/pull/98/files#diff-ccb5c6ab1b4e9c6892cba9f2a904889b49d1e52f336bd2d558ff07ec9ffc5aa2R17

packages/gatsby-theme-carbon/src/components/LeftNav/LeftNavTree.module.scss
elycheea: @szinta brought up that `$layer` doesn’t map quite right for the themes — we need `white` in light theme but `#161616` in dark theme, which only `$text-inverse` seems to match.

She will follow the approach from regular left nav and add custom styles here.

packages/gatsby-theme-carbon/src/components/LeftNav/LeftNavTree.module.scss
szinta: gatsby-theme-carbon doesn't use logical properties, hence added border using `border-right` prop

packages/gatsby-theme-carbon/src/components/LeftNav/LeftNavTree.module.scss
szinta: used $background token for dark theme and added custom style overrides for g10 theme

packages/example/src/data/nav-items.yaml
```diff
@@ -48,6 +48,7 @@
       path: /guides/v3-migration
     - title: v4.0 Migration
       path: /guides/v4-migration
+  hasDivider: true
```
alisonjoseph, line 51: Was this added on purpose?

alisonjoseph (commented):
this LGTM other than the question about the extra nav divider

packages/example/src/data/nav-items.yaml
szinta: I added it for testing purpose only. I have now removed it.

alisonjoseph (approved):

elycheea (approved):
🎉

PR #1523, opened 2024-09-30
fix: caption spacing inside tab

### LINKED ISSUE #1493: [Image caption] Add captions under each tab image
## Detailed description
There are a couple of issues identified:

1. Captions: We need to add individual captions under each tab image. When a caption is placed inside a tab, there is a missing 1rem top padding above the caption.
2. Background Display: The white background observed is the tab container itself, not a background on the image. As a result, anything inside the tab will display on that white background.

While we could build a custom solution to remove the white background, the current implementation is functioning as intended, aside from the missing 1rem top padding.

image

## Possible solution 
(@alisonjoseph mentioned -)
To address these issues, we need to enhance the Gatsby tabs component. A potential solution could involve creating a new tab type, such as `<Tab type="image">`, allowing us to style this variant differently to remove the background and adjust padding as needed.

## Additional information
- [Related issue](https://github.com/carbon-design-system/gatsby-theme-carbon/issues/1488)
- Ref: [Thread](https://ibm-studios.slack.com/archives/GSCEH4RSP/p1715693675697239?thread_ts=1715688078.860809&cid=GSCEH4RSP)

alisonjoseph:
Closes #1493

#### Changelog
- add correct top margin above captions

## Testing
Check captions on Tabs and Captions page. Should have 1rem top margin.

### DIFF
--- a/packages/example/src/pages/components/Tabs.mdx
+++ b/packages/example/src/pages/components/Tabs.mdx
@@ -28,7 +28,9 @@ columns. If you want to have multiple columns, you’ll need to use our `Row` an
 
 If your content is full width, you don’t need `Row` and `Column`
 
-![color array](images/colors.jpg)
+![color array](./images/colors.jpg)
+
+<Caption>This is a caption</Caption>
 
 </Tab>
 
--- a/packages/gatsby-theme-carbon/src/components/Caption/Caption.module.scss
+++ b/packages/gatsby-theme-carbon/src/components/Caption/Caption.module.scss
@@ -1,6 +1,7 @@
 .caption {
   @include type-style('body-short-01');
   --space: #{$spacing-05};
+  margin-top: var(--space);
 }
 
 div .caption * {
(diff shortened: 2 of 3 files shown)


Kritvi-bhatia17 (commented):
Hi Alison! 
Will the caption appear within the image, meaning will the caption have the same white background as the image? 🤔

image

--- 

(This is what I’m looking at 👇🏻, I hope that’s correct)

image

alisonjoseph:
@Kritvi-bhatia17 the white background is coming from the Tabs component, not the image. The caption is within the tab, so it will appear on the white background. The only thing this PR is doing is adding the correct spacing above the caption that was missing when it was used inside of the Tabs component.

ariellalgilmore (approved):

Kritvi-bhatia17:
Oh, got it, Alison. Thanks a lot for the detailed explanation! 
In that case, I’ll just confirm with the other designers as well and will get back to you.

Kritvi-bhatia17:
Hi @alisonjoseph, I discussed with the team, and it seems a bit inconsistent with other images where the caption appears below the image. 
Initially, I thought adding captions for the tab images would improve consistency, but given the current solution, it might look a bit odd and cause confusion, especially with images (shown below) that already include text. It could become unclear which is the caption and which is part of the artwork. 
Sorry about this and thanks for working on the issue! I just wanted to explore the possibility of adding captions for the tab images below them.

image

PR #1578, opened 2025-03-24
fix: allow multiple medium accounts

preetibansalui:
Closes #4476 from Carbon-Website

Updated the gatsby-config function to support multiple Medium accounts by accepting an array of strings as well.


alisonjoseph:
Looks good, can we update the docs to show an example of using multiple accounts? https://github.com/carbon-design-system/gatsby-theme-carbon/blob/main/packages/example/src/pages/components/MediumPosts.mdx

PR #1659, opened 2025-12-11
fix(globalsearch): prevent page crash when typing `:` or `^`

maradwan26:
No issue

Fixes a bug where `GlobalSearch` would crash the page when the user types `:` or `^`.

This was due to Lunr interpreting the characters as part of its query syntax, hence the `QueryParseError`: `"unrecognised field '', possible fields: 'title', 'body'"` when typing `:` and `"expected either a field or a term, found BOOST"` when typing `^`. 

I added a function that escapes special characters to prevent this. I've also tested regular search queries for regressions to ensure this change does not affect search results.

#### Changelog

**New**

- Added a `sanitizeQuery` function that uses a regex to escape special characters

#### Testing and Reviewing
- Typing `:` and `^` into the search bar at the top should not crash the page
- Feel free to test this[ alongside the current `GlobalSearch`](https://gatsby.carbondesignsystem.com/) to ensure there aren't any regressions to the search results

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 # Changelog
 
+## [4.3.24](///compare/v4.3.23...v4.3.24) (2025-12-10)
+
+## [4.3.23](///compare/v4.3.22...v4.3.23) (2025-12-10)
+
+### Bug Fixes
+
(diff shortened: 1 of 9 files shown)


alisonjoseph (approved):
🚀 LGTM

